PR #18, opened 2018-10-27
Add boxnovel.com as new source

yudilee:
I add new source for boxnovel.com.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,9 @@ The avaiable list of site handles are given below. *To request new site [create 
 | Handle | Website |
 |--------|---------|
 | `webnovel` | https://www.webnovel.com |
-| `wuxia` | http://www.wuxiaworld.com |
-| `wuxiac` | http://www.wuxiaworld.co |
+| `wuxia` | https://www.wuxiaworld.com |
+| `wuxiac` | https://www.wuxiaworld.co |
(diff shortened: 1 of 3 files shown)


dipu-bd:
Good to see a new source! It pains me that I have not been getting any free time lately to invest on this project. I have uploaded a new version to *pypi* merging this pull request: https://pypi.org/project/ebook-crawler/

yudilee:
Thanks a lot for merge this pull request. I have already have new
source(novelplanet) but I have to use cfscrape to get the source code
instead of using requests because of usage of antibot on this site. May I
add pull request on this code because it will add new requirement for this
app. Please gave me suggestion on this. Should I pull request first or do
you have any workaround on this matter.

Best Regard

Yudi Lee

dipu-bd:
Cloudflare is tricky. It checks javascript support. `cfscrape` uses node.js to bypass the anti-bot check. So, user must have node.js installed in their PC.

I do not think this is a problem. If someone wants to download particularly from novelplanent, they should be prompt to have `node.js` installed.

Make your pull request to the `dev` branch. I have to review it before merging with `master`.

yudilee:
yes, user should install node.js in their pc to use it. in also use it in binding file to get cover page for the novel. i will try to create branch and upload it. thanks for your reply.

PR #19, opened 2018-11-07
Add New Source novelplanet and add cfscrape in requirement

yudilee:
Add New Source novelplanet and add cfscrape in requirement


dipu-bd:
I moved the cover image retrieval method to `helper.py`. first it will try to use `cfscrape` and if fails will use old style `requests`

yudilee:
Ok. Thanks. Really nice idea.

PR #23, opened 2018-11-09
Convert old crawlers to new style

yudilee:
I have convert old crawler to new template. Adding some change in crawler class to enabling using cfscrapper for novelplanet. Fix Helper to enable it get cover using cfscrapper (it happen to error before). Fix url link for novel cover in wuxia crawler.


dipu-bd:
Great! You reduced lot of my work :smile_cat:

yudilee:
Thank you

PR #57, opened 2019-01-29
Add option to choose book format and minor fix in boxnovel crawler

yudilee:
Add option to choose book format and minor fix in boxnovel crawler


dipu-bd:
Considering the console version, the binding should be independent of compressing. I changed reflecting this ideology.

yudilee:
can we not compress epub, pdf, docx and mobi file? so it's easier for user to use bot to get book in that format, not decompressing the file to access that book

dipu-bd:
I compressed the output because there could be multiple files. Additionally it reduces the file size.

But if there is a single output file, we do not have to compress it. I thought about it and decided to add some extra conditions. But have not implemented it yet.

I will do something like this: Inside `core/app.py`, when compressing the file, I will first get the file count inside the folder. If there is only a single file I will add that file to the `archived_outputs` variable. After the latest push all files inside of archived_outputs will be uploaded and sent to user.

yudilee:
Agree on this. I also want to add ability to telegram bot to select range of chapter. Still working on this, I will pull request as soon as I get it work.

PR #90, opened 2019-02-25
Add new crawler and minor fix to 3 crawler

yudilee:
Add new crawler for Indonesian translation novel source : meionovel.com
Minor fix to RoyalRoad, Scribblehub and Novelplanet crawler to enable get table inside chapter (usually for novel with game genre that has stat described using table)


yudilee:
After some testing, adding table to novel chapter create problem in pdf generator. All pdf output generates blank chapter

dipu-bd:
Awesome! We have got 31 sources now! I could not imagine there would be so many sources someday

yudilee:
Really happy with this program structure you designed that made it possible for me to contribute to this project.

PR #98, opened 2019-03-07
Add site litnet

drewbitt:
Litnet: Site of web novel publishing, has some litrpg/fantasy series like RR

Had an interesting time with this. Found out that 1/10th of the time search does not give you english results, so you get /book/ and /en/book/ results, but even going to /en/book/ doesn't give you the same html as before. Likely the language cookie being set differently but didn't look into it. So had to write two parsers for the different tags.


dipu-bd:
Thanks for the pull request. I was away for a while. Merging it was delayed as a result.

PR #108, opened 2019-03-13
Update search method in 2 crawler and minor fix in telegram bot

yudilee:
Update search method in 2 crawler and minor fix in telegram bot


dipu-bd:
I merged your pull request. However, I changed one thing. In both crawler files, there was no need to create separate instance of cfscrape. I used the one from the base `Crawler` class.

yudilee:
Thank you, thank for inform me. sorry coud't find it before

PR #161, opened 2019-07-04
Fix discord.py and python-telegram-bot version

yudilee:
Fix discord.py and python-telegram-bot version


dipu-bd:
how did I miss it while reviewing ¯\_(ツ)_/¯

PR #207, opened 2019-10-11
Heroku deployment support

NNTin:
Log Preview: 
image

Go here: https://github.com/NNTin/lightnovel-crawler/tree/heroku  
In the ReadMe click on the Heroku Button: 
image

Fill out the Discord Bot token and hit deploy.
Analog example with another project: https://www.youtube.com/watch?v=NwPcXBvStSI  

Maybe breaking changes. You'd have to decide:  
I explicitely locked the requirements.txt and removed requirements_dev.txt

```bash
python3.6 -m venv venv
source venv/bin/activate
python -m pip install -r requirements.txt
python -m pip install -r requirements_dev.txt
rm requirements_dev.txt
python -m pip freeze > requirements.txt
```

Reason:
When you deploy on Heroku a Python App it looks for runtime.txt in which the Python version is specified. It further looks for requirements.txt. Since I don't know which dependencies are "directly" and "indirectly" required I took the autogenerated one.


dipu-bd:
I changed `requirements.txt` however. I try to keep this list minimum to decrease EXE bundle size. Now I added enough packages to support bots, it should not have any issue with Heroku.

Issue #282, opened 2019-12-24
[gravitytales] Fix novel info parsing


dipu-bd:
**Describe the bug**
The novel has 853 chapters, but app shows 1706

![image](https://user-images.githubusercontent.com/5158124/71421656-6ccca900-26a6-11ea-8e14-ab86bfe5e86d.png)

**Let us know**
- Novel URL: http://gravitytales.com/novel/shuras-wrath
- App Version: 2.17.1
- Package: pip

diogenes895:
https://github.com/dipu-bd/lightnovel-crawler/pull/290 this should fix it

PR #290, opened 2020-01-01
fix gravitytales parsing bug 282

diogenes895:
bug fix for https://github.com/dipu-bd/lightnovel-crawler/issues/282


dipu-bd:
thanks for the fix. i’ll merge it tomorrow

PR #327, opened 2020-02-03
Added scraper for Boxnovel.org with search capability.

PreownedFIN:
Added a single test input to bots' test_inputs.py.

I tested with multiple different novels on the site (with search and direct urls), and with cursory glancing I didn't notice any formatting problems with the 'plain' formatting.

Additional notes:
To differenciate between boxnovel.**_com_** and boxnovel.**_org_**, I named the org scraper as "boxnovelorg". If there is a better fitting name, it should/could be changed.

Sidenote: Salty Panda recommended a webnovel that wasn't on any site besides boxnovel.org (or any that lncrawl search was working for), so I did what I had to do, to get it for offline reading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,7 @@ The list of currently available sources are given below:
 | https://bestlightnovel.com             |     ✔      |           |
 | https://book.qidian.com                |            |           |
 | https://boxnovel.com                   |     ✔      |           |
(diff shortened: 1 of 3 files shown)


dipu-bd:
Thanks @PreownedFIN 
Your code has been merged.

PR #365, opened 2020-03-09
Replace cfscrape with cloudflare

dipu-bd:
- [v3] Replace cfscrape with cloudflare
- [mtled] disable search for cloudflare issue


dipu-bd:
fixes #358

PR #481, opened 2020-06-08
fix: [novelfull] use match and non-capturing group

### LINKED ISSUE #459: Crawler failing on specific chapter on novelfull.com
Crawler systematically fail on chapter 111 of https://novelfull.com/reborn-evolving-from-nothing.html

**Let us know**
- Novel URL: https://novelfull.com/reborn-evolving-from-nothing.html
- App Version: 2.21.0
- Package: exe
- OS: Windows 10

image

kuwoyuki:
Use `re.search` instead of `re.findall`, add a non-capturing group, compile regex
outside of the loop, fixes #459

### DIFF
--- a/src/sources/novelfull.py
+++ b/src/sources/novelfull.py
@@ -6,7 +6,9 @@ from ..utils.crawler import Crawler
 
 logger = logging.getLogger('NOVEL_FULL')
 search_url = 'https://novelfull.com/search?keyword=%s'
-
+# avoid compiling regex inside the loop
+RE_CHAPTER = r'(?:ch(apter))? (\d+)'
+RE_VOLUME = r'(?:book|vol|volume) (\d+)'
 
 class NovelFullCrawler(Crawler):
     base_url = [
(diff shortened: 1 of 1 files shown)


src/sources/novelfull.py
```diff
@@ -91,19 +93,20 @@ def download_chapter_list(self, page):
...
-                               title, re.IGNORECASE)
-            if len(match) == 1:
-                volume_id = int(match[0][1])
-            # end if
+            if ch_match:
+                chapter_id = int(ch_match.group(2))
+
+            _v_id = vol_match.group(1) if vol_match else None
+            if _v_id and (
+                not prev_ch_vol or _v_id == prev_ch_vol or _v_id == prev_ch_vol + 1
```
kuwoyuki, line 107: this uses the matched index if `self.chapters` is empty or if the matched vol is the same or +1 as previous `chapters.volume`

dipu-bd: nicely done

PR #549, opened 2020-08-13
New source

SirGryphin:
I've added fastnovel.py source and tested it, seems to work okay. Also edit readme.md to added it to supported sources. 

New to python and github so if I've done this wrong please let me know. Some of my commits are Gryphin or SirGryphin because it's first time using VS Code and Git and I had a bit of a problem with setting, I hope that's okay as I noticed they don't have the "Verified" tag.

Also I didn't know if I was meant to send request to dipu-bd:dev or dipu-bd:master, or doesn't it matter could you let me know for future reference.


dipu-bd:
Thanks!

SirGryphin:
Okay thanks 👍

PR #554, opened 2020-08-14
Minor Fix & New Source

SirGryphin:
I noticed when opening the book that there is a double line/space between paragraphs because of the `<br><br>` tags. This bit of code removes them so the books are properly formatted.

**Update:** Just created source file for Indonovels.net


dipu-bd:
thanks

PR #558, opened 2020-08-17
Add Source

SirGryphin:
**https://fujitranslation.com/**


dipu-bd:
When adding a crawler, please also update README.md file. Add the crawler links at the end of the [Supported sources](https://github.com/dipu-bd/lightnovel-crawler/tree/dev#c3-supported-sources).

SirGryphin:
I will do from now on, sorry forgot about that.

PR #570, opened 2020-08-25
Source Added

SirGryphin:
Added  https://www.fuyuneko.org/

Also edit readme and added source and remove duplicate/wrong link for dmtranslationscn.com/wp there should be no /wp.


dipu-bd (approved):

dipu-bd:
thanks a lot @SirGryphin

PR #599, opened 2020-09-07
minor fixes

SirGryphin:
Add search feature to a few sources and fix a cover image problem.

Wuxiaworld.co and Readlightnovel.cc search fixes are thanks to Yudi Lee, they were to busy to make pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,7 @@ You are very welcome to contribute in this project. You can:
 | https://viewnovel.net                        |     ✔      |           |
 | https://vipnovel.com                         |     ✔      |           |
 | https://volarenovels.com                     |            |           |
-| https://wbnovel.com                          |            |           |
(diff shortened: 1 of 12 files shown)


dipu-bd:
Thanks for your pull request. I have some confusions however.

You have added some comments saying search is not working properly. If a search method does not work, you should keep it commented out.

About `lncrawl/sources/wuxiasite.py` search: it was disabled because the site uses cloudflare, and enabling this search increases overall search time by a lot.

Since there are already a lot of sources with search feature, we should not enable searching for those sources that are slow or does not work properly. It will affect user experience badly.

SirGryphin:
Commented out search code that doesn't work like asked and also commented out wuxia site like as because of cloudflare. Hope that sorts everything.

dipu-bd:
its ok now. Thanks again @SirGryphin

PR #604, opened 2020-09-11
Update creativenovels.py

tomcb1:
Added another for loop to download_chapter_body() that removes the div element with the class: "novel_showcase" which leaves the content of body only with the content of the novel.


dipu-bd:
thanks for this PR

PR #609, opened 2020-09-17
Fixed an issue where some chapters would just be left blank.

tomcb1:
With [this chapter](https://creativenovels.com/80/volume-6-chapter-22-bits-of-the-answer/) I noised that text was contained within span tags. So I made a change where span tags with a transparent style such as: `<span style="color:transparent;">` Are removed before the remaining span tags are replaced with `<p>`.


dipu-bd:
thanks for the PR

PR #614, opened 2020-10-01
Minor fixes to novelfull

SirGryphin:
Fixed missing brackets at the end of a decompose. Made it so it removes junk text and double chapter heading. I've added comments to file to explain a bit better.

### DIFF
--- a/lncrawl/sources/novelfull.py
+++ b/lncrawl/sources/novelfull.py
@@ -134,21 +134,23 @@ class NovelFullCrawler(Crawler):
             r'Read more chapter on NovelFull'
             r'^full thich ung',
         ]
+        
+        bad_class = ['h3', 'h2', '.adsbygoogle', 'script', 'ins', '.ads', '.ads-holder']
+        def is_ad(tag):
+            return isinstance(tag, Comment) or (
+                tag.has_attr('class') and \
+                any(x in tag.get("class") for x in bad_class)
+            )
+        for ads in content.find_all(is_ad):
+            ads.decompose()
 
         # Some comments in source code of site gets converted into text/paragraphs. This removes it.
-        for comment in soup.findAll(text=lambda text:isinstance(text, Comment)):
-            def is_ad(tag):
-                return isinstance(tag, Comment) or (
-                    tag.name == "div" and tag.get("class") in [["left"], ["center"]]
-                )
-            for ads in content.find_all(is_ad):
-                ads.decompose()
-            comment.extract()
+        #for comment in soup.find_all(text=lambda text:isinstance(text, Comment)):
+        #    comment.extract()
 
-        self.clean_contents(content)
-
-        for ads in content.select('h3, h2, .adsbygoogle, script, ins, .ads, .ads-holder'):
-            ads.decompose()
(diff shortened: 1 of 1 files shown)


lncrawl/sources/novelfull.py
```diff
@@ -122,14 +123,33 @@ def download_chapter_body(self, chapter):
...
+        self.blacklist_patterns = [
+            r'^Translator:',
+            r'^Editor:',
+            r'^Atlas Studios',
+            r'Read more chapter on NovelFull'
+            r'^full thich ung',
+        ]
+
+        # Some comments in source code of site gets converted into text/paragraphs. This removes it.
+        for comment in soup.findAll(text=lambda text:isinstance(text, Comment)):
```
kuwoyuki, line 139: Since you're adding a custom function here, could you move the below `findAll`s and `select`s here?
```py
def is_ad(tag):
    return isinstance(tag, Comment) or (
        tag.name == "div" and tag.get("class") in [["left"], ["center"]]
    )
for ads in content.find_all(is_ad):
    ads.decompose()
```
Also please use `find_all` https://www.python.org/dev/peps/pep-0008/

kuwoyuki, line 139: Since you're adding a custom function here, could you move the below `findAll`s and `select`s here?
```py
def is_ad(tag):
    return isinstance(tag, Comment) or (
        tag.name == "div" and tag.get("class") in [["left"], ["center"]]
    )
for ads in content.find_all(is_ad):
    ads.decompose()
```
Also please use `find_all` https://www.python.org/dev/peps/pep-0008/

kuwoyuki (changes requested):

lncrawl/sources/novelfull.py
SirGryphin: I'm very new to python, most of my knowledge is from reading other people code and seeing how it works, or googling a problem I have and seeing if someone has got some code I can change slightly to fix the problem. The code you supplied is something I couldn't write from scratch and above my understanding for now. I'm guessing it works so thank you for the changes made.

dipu-bd: @SirGryphin please add the changes suggested by @kuwoyuki

SirGryphin: Added suggested code and tested it, everything works. Hopefully this is what you meant.

dipu-bd:
@SirGryphin Your edit was not exactly what @kuwoyuki was talking about. I modified it a little and made another push. Please test if it is working

SirGryphin:
Doesn't work, I tested with the URL below, If you download last 10 chapters you will see the comment "full thich ung" still shows as text in chapter. The only I've found to remove comment is with the following code using extract():

```
for comment in soup.find_all(text=lambda text:isinstance(text, Comment)):
            comment.extract()
```

https://novelfull.com/priceless-new-bride-accidentally-offending-the-dangerous-ceo.html

dipu-bd:
@SirGryphin have you pulled the latest dev branch?

SirGryphin:
Ahh, my fault mine was slightly outdated. Seems to be working fine now thank you.

Issue #344, opened 2020-02-22
https://foxaholic.com/


SirGryphin:

SirGryphin:
Is there anyway to make it so images download as well and don't show up as error boxes? One the novels has a lot of illustrations and I want them in book if possible.

PR #670, opened 2020-11-14
Added support for foxaholic.com

frybin:
Solves Issue  #344


dipu-bd:
Thanks a lot! I was planning to add it for a while, but could not make time.

frybin:
No worries, it was mainly like wbnovel.com with them both using the same wordpress theme though loading the ToC needed an additional post request which was not as fun.

PR #717, opened 2021-01-11
Added scraper for readlightnovels.net with search capability.

PreownedFIN:
Added single test input to bots' test_inputs.py for the site.


PreownedFIN:
Hold up, they do the pagination in a funky way, so still needs a bit of work.

PreownedFIN:
If you check the branch (26c103b), I added an import to bs4, because the Crawler class 'builtins' couldn't handle an URL that responds with HTML inside JSON.
The other way would have been to add a new function to crawler.py, but I decided to stay in the one file.

Also, the pagination calls **an arbitary URL** that's defined on line 96, it is the only place you can get the pagination results, **so I hardcoded it**.
It could be read from the javascript of the page, but it seems to work fine as-is.

If the bs4 functions need to be defined in the Crawler class (none in the scrapers extending it), I can refactor the code to be so.
Also the hardcoded URL is kinda sketch...

dipu-bd:
@PreownedFIN I have reviewed your code, all seems fine. For arbitrary URLs, I usually put them above the crawler class. But defining them inside the method is also okay. For the URLs that won't change often, we can keep them hard-coded.

There is a slight issue however. You are pushing directly to `master` branch. Can you change it to `dev` branch? Let me know if you can do this easily.

PreownedFIN:
Alright, I changed the destination branch to your dev branch in the 'edit pull request' widget that Github has.
I resolved the conflict in the README.md, and added the readlightnovels.net as a supported source.

dipu-bd:
@PreownedFIN Thanks!

Issue #724, opened 2021-01-18
Babel Novel not working


PleaseReadMe:
**Describe the bug**

**Let us know**
- Novel URL: [https://babelnovel.com/books/cultivation-from-cellphone]
- App Version: [2.24.1]
- Package: [pip]
- OS: [e.g. Android 8.1.0]
![Screenshot_2021-01-18-23-52-01-97](https://user-images.githubusercontent.com/70464689/104936992-3b78f680-59e8-11eb-9a24-4f24e199b64d.jpg)

PR #730, opened 2021-01-25
Fix downloading from babelnovel.com

Galunid:
* Remove parse_content_css function that didn't work and was breaking downloading chapter body
* Update api urls

fixes https://github.com/dipu-bd/lightnovel-crawler/issues/724

### DIFF
--- a/lncrawl/sources/babelnovel.py
+++ b/lncrawl/sources/babelnovel.py
@@ -12,10 +12,10 @@ from ..utils.crawler import Crawler
 
 logger = logging.getLogger(__name__)
 
-search_url = 'https://babelnovel.com/api/books?page=0&pageSize=8&fields=id,name,canonicalName,lastChapter&ignoreStatus=false&query=%s'
-novel_page_url = 'https://babelnovel.com/api/books/%s'
-chapter_list_url = 'https://babelnovel.com/api/books/%s/chapters?bookId=%s&page=%d&pageSize=100&fields=id,name,canonicalName,hasContent,isBought,isFree,isLimitFree'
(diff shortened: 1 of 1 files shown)


lncrawl/sources/babelnovel.py
```diff
@@ -119,35 +116,11 @@ def parse_chapter_item(self, list_url):
...
-    # end def
-
     def download_chapter_body(self, chapter):
         logger.info('Visiting %s', chapter['json_url'])
         data = self.get_json(chapter['json_url'])
 
         soup = BeautifulSoup(data['data']['content'], 'lxml')
-        if self.bad_selectors:
-            for tag in soup.select(self.bad_selectors):
-                tag.extract()
```
dipu-bd, line 147: babel novel puts some jumbled text along with main chapter content, which are hidden using CSS. if you remove this, the jumbled text will be visible in the epub.

dipu-bd (changes requested):

lncrawl/sources/babelnovel.py
Galunid: I don't think they do that anymore. I downloaded 10 chapters of random novel and didn't see any weird strings. Then I browsed the HTML of a different one and there were no strange things in the chapter content HTML. Are you sure it's still a thing? As it is it doesn't work either way, since the regex doesn't match anything and the whole thing just throws an exception.

dipu-bd: yeah, you are right. i've checked several novels now, it doesn't do that anymore.

PR #752, opened 2021-02-11
Sources

SirGryphin:
**Added:** 
https://morenovel.net
https://dsrealmtranslations.com
https://hui3r.wordpress.com

~~Hui3r, source works okay but there is a bit of "Junk" text that I'm having problem getting rid of at bottom of chapters. It was a choice between two ways of getting chapter body, one leave copyright watermark and other leave junk at bottom.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,7 @@ You are very welcome to contribute in this project. You can:
 | https://mangatoon.mobi                       |            |           |                       |
 | https://meionovel.id                         |     ✔      |           |                       |
 | https://moonstonetranslation.com             |            |           |      @SirGryphin      |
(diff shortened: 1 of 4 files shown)


Galunid:
Hui3r seems to put all the junk in `<span><em></em></span>` tags, so perhaps you can just extract them? So one of the patterns may looks something like
`r'<p><span .*?><em>.*?</em></span></p>'`

Which hopefully matches
```html
<p><span style="color: rgb(255, 255, 255);"><em>Junk...</em></span></p>
```

SirGryphin:
@Galunid thanks for the suggestion. I've ended up re-doing the whole thing and it seems to work better now all junk text is code and no hidden watermark either.

PR #757, opened 2021-02-13
Update crawler.py

BorgSquared:
Fixes issues with many sources when searching.

e.g.: 
With the string slice you get: https://www.mywuxiaworld.combook/Nanomancer_Reborn_-_I_ve_Become_A_Snow_Girl/
Without: https://www.mywuxiaworld.com/book/Nanomancer_Reborn_-_I_ve_Become_A_Snow_Girl/

### DIFF
--- a/lncrawl/utils/crawler.py
+++ b/lncrawl/utils/crawler.py
@@ -141,7 +141,7 @@ class Crawler:
         elif url.find('//') >= 0:
             return url
         elif url.startswith('/'):
-            return self.home_url + url[1:]
+            return self.home_url + url
         elif page_url:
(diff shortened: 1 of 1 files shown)


Galunid:
I was about to submit the same patch, explanation why it's needed:
https://github.com/dipu-bd/lightnovel-crawler/blob/8a9b5f0bfdbd38a53179ddc2088a8dcfbb601a34/lncrawl/core/app.py#L105
Strips the "/" from the `self.home_url`, so if we strip `/` from `url`, we get wrong result

dipu-bd:
thanks a lot

PR #768, opened 2021-02-20
Use shortest match for replace (novelfull)

Galunid:
Avoids situations when whole chapter body is replaced with nothing
ex. `<p>chaptercontent</p><p>Translator:</p>`

This shouldn't happen often, only when Translator/Editor notices are at the end of the chapter, not the beginning. Out of 40+ novels I tested, it happened once


dipu-bd:
thanks for the fix

Issue #813, opened 2021-04-19
Fix this bug: Heroku deployment not working


krispykiq:
I tried using the deploy from heroku button and filled out all the needed envinroment variables. 

It said:

```
-----> Build succeeded!
-----> Python app detected
 !     Requested runtime (python-3.6.9) is not available for this stack (heroku-20).
 !     Aborting.  More info: https://devcenter.heroku.com/articles/python-support
 !     Push rejected, failed to compile Python app.
 !     Push failed
```

So I tried fixing it by cloning the repository and changing the python version in runtime.txt to python-3.6.13 but then another error appeared saying

```
ar: libnss3_3.58-1_amd64.deb: file format not recognized
 !     Push rejected, failed to compile Calibre app.
 !     Push failed
```

damare01:
please fix it!

dipu-bd:
Have you tried using python 3.8?

damare01:
> Have you tried using python 3.8?

how?

damare01:
how do I change my pyhton runtime version? Mine is 3.9.4

damare01:
> > Have you tried using python 3.8?
> 
> how?

never mind

damare01:
> how do I change my pyhton runtime version? Mine is 3.9.4

never mind

damare01:
> Have you tried using python 3.8?

won't work

damare01:
this is what I get

damare01:
this is what I get 

http://ftp.debian.org/debian/pool/main/n/nss/libnss3_3.58-1_amd64.deb
ar: libnss3_3.58-1_amd64.deb: file format not recognized
 !     Push rejected, failed to compile Calibre app.
 !     Push failed

damare01:
this is what I get

damare01:
this is what I get 

http://ftp.debian.org/debian/pool/main/n/nss/libnss3_3.58-1_amd64.deb
ar: libnss3_3.58-1_amd64.deb: file format not recognized
 !     Push rejected, failed to compile Calibre app.
 !     Push failed

damare01:
tried this versions - **python-3.9.4 python-3.8.9 python-3.7.10 python-3.6.13** it doesn't work. I always get this "http://mirrors.kernel.org/ubuntu/pool/main/libg/libglvnd/libgl1_1.0.0-2ubuntu2_amd64.deb
[...]

PR #832, opened 2021-04-26
Update Heroku build

NNTin:
Related to issue https://github.com/dipu-bd/lightnovel-crawler/issues/813

Removed `runtime.txt`: Current the default supported Python version is 3.9.4. Tested with it and it works: https://devcenter.heroku.com/articles/python-support#specifying-a-python-version  
You can specify a fixed version but you'll have to keep the file updated.

In the calibre buildpack I removed the dependencies since they are tied to a version and may become no longer available: https://github.com/NNTin/heroku-buildpack-calibre/commit/443a77d105a4a919e9422195cade76b18a219a3b

The dependencies are moved instead into the Aptfile: https://github.com/NNTin/lightnovel-crawler/blob/dev/Aptfile

**Tested the discord bot and it supports epub text web mobi fb2 but not pdf.**

---

For pdf support `python3-pyqt5.qtwebengine` needs to be added to `Aptfile` or `PyQtWebEngine` needs to be added to `requirements.txt` to fix the error:

`ImportError: cannot import name 'QWebEngineUrlScheme' from 'qt.webengine'`

mentioned here: https://github.com/dipu-bd/lightnovel-crawler/issues/813#issuecomment-824858181

However when doing so you'll get a different error saying the slug size is too big:

`Compiled slug size: 690.7M is too large (max is 500M).` --> modifying Aptfile
`Compiled slug size: 515.2M is too large (max is 500M).` --> modifying requirements.txt

When slug size exceeds 500M your Heroku dyno won't deploy at all.

Currently when deploying the slug size is:

`Warning: Your slug size (386 MB) exceeds our soft limit (300 MB) which may affect boot time.`


NNTin:
PDF now works.
Apparently Calibre is bundled with a few libraries already.
These libraries needed to be added to `LIBRARY_PATH`: https://github.com/NNTin/heroku-buildpack-calibre/blob/5be5fde94901af4e7053482c6673d0ef7eab7dc8/bin/compile#L33

See:
image

but a few of them are still missing. I readjusted `Aptfile`. To install the missing dependencies.
Slug size is at 381 MB.

dipu-bd (approved):

PR #869, opened 2021-05-15
Fix: sources/1stkissnovel.py

mchubby:
Based on fix for source/noveltranslate.py. Adds search.


mchubby:
I copied them dumbly but I think the re and urllib imports are not required

mchubby:
@SirGryphin If the PR is okay with you, the readme can mark this source as searchable.

SirGryphin:
@mchubby thanks for fixing search, also the readme already has it marked as "searchable". The only thing you can do is update the readme.

**From this:**
`| https://1stkissnovel.love                    |     ✔      |           |      @SirGryphin      |`

**To this:**
`| https://1stkissnovel.love                    |     ✔      |           |@mchubby, @SirGryphin|`

mchubby:
It's fine; I don't really mind not being listed as one of the maintainers :)
The source being very similar to other ones based on the same plugin like skynovel etc., it should be relatively easy to fix should it break.

lncrawl/sources/1stkissnovel.py
```diff
@@ -1,81 +1,104 @@
...
     def download_chapter_body(self, chapter):
-        '''Download body of a single chapter and return as clean html format.'''
-        logger.info('Downloading %s', chapter['url'])
-        soup = self.get_soup(chapter['url'])
+        """Download body of a single chapter and return as clean html format."""
+        logger.info("Visiting %s", chapter["url"])
+        soup = self.get_soup(chapter["url"])
 
-        contents = soup.select_one('div.text-left')
+        contents = soup.select_one('div.text-left div:first-child')
```
dipu-bd, line 101: this does not work on many chapters. e.g.: 

- https://1stkissnovel.love/novel/the-reborn-wife-is-too-seductive/chapter-9/
- https://1stkissnovel.love/novel/the-reborn-wife-is-too-seductive/chapter-10/

dipu-bd (changes requested):

lncrawl/sources/1stkissnovel.py
mchubby: Good catch!

PR #908, opened 2021-06-07
Dev

SirGryphin:
Removed `@cleanup_text` from foxaholic.py because it was causing this error to pop up `! Error: Src` 

**Added:**
https://isotls.com
https://lazybirdtranslations.wordpress.com
https://readnovelz.net
https://www.novelmultiverse.com

**Note:** Sorry for weird list of commits below, I did a `git rebase upstream/dev` command instead of my normal `git merge upstream/dev` and it seems to of listed them twice, seem everything else is okay though no files changed etc,

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,6 +463,7 @@ You are very welcome to contribute in this project. You can:
 | https://www.novelcool.com                    |            |           |      @SirGryphin      |
 | https://www.novelhall.com                    |            |           |                       |
 | https://www.novelhunters.com                 |     ✔      |           |      @SirGryphin      |
+| https://www.novelmultiverse.com              |     ✔      |           |      @SirGryphin      |
 | https://www.novelpassion.com                 |     ✔      |           |      @SirGryphin      |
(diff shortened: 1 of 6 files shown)


dipu-bd:
`@cleanup_text` was removing unwanter texts, ads etc from foxaholic. If you just remove this, the chapter content of foxaholic might get gibberish. I would rather fix the cleanup_text method instead of removing it.

If you don't feel comfortable to fix the method, you can leave an issue open. I will check it on my next free schedule.

dipu-bd (changes requested):
lncrawl/sources/foxaholic.py 

do not remove `@cleanup_text`, fix the method instead

SirGryphin:
I found it was a lazy loading img causing the error on certain books, I've added some code to remove the lazy attribute from image to fix error. Is this code okay now?

dipu-bd:
> I found it was a lazy loading img causing the error on certain books, I've added some code to remove the lazy attribute from image to fix error. Is this code okay now?

Looks okay. Thanks

Issue #1052, opened 2021-08-28
Foxaholic, Body is empty


watzeedzad:
**Describe the bug**

Got Body is empty error when tried to download on foxaholic

```
lncrawl                               
================================================================================
╭╮╱╱╱╱╱╱╭╮╱╭╮╱╱╱╱╱╱╱╱╱╱╱╱╭╮╱╭━━━╮╱╱╱╱╱╱╱╱╱╭╮
┃┃╱╱╱╱╱╱┃┃╭╯╰╮╱╱╱╱╱╱╱╱╱╱╱┃┃╱┃╭━╮┃╱╱╱╱╱╱╱╱╱┃┃
┃┃╱╱╭┳━━┫╰┻╮╭╋━╮╭━━┳╮╭┳━━┫┃╱┃┃╱╰╋━┳━━┳╮╭╮╭┫┃╭━━┳━╮
┃┃╱╭╋┫╭╮┃╭╮┃┃┃╭╮┫╭╮┃╰╯┃┃━┫┃╱┃┃╱╭┫╭┫╭╮┃╰╯╰╯┃┃┃┃━┫╭╯
┃╰━╯┃┃╰╯┃┃┃┃╰┫┃┃┃╰╯┣╮╭┫┃━┫╰╮┃╰━╯┃┃┃╭╮┣╮╭╮╭┫╰┫┃━┫┃
╰━━━┻┻━╮┣╯╰┻━┻╯╰┻━━╯╰╯╰━━┻━╯╰━━━┻╯╰╯╰╯╰╯╰╯╰━┻━━┻╯
╱╱╱╱╱╭━╯┃ v2.28.3
╱╱╱╱╱╰━━╯ 🔗 https://github.com/dipu-bd/lightnovel-crawler
--------------------------------------------------------------------------------

➡ Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://www.foxaholic.com/novel/i-was-caught-up-in-a-hero-summoning-but-that-world-is-at-peace/  
Retrieving novel info...
https://www.foxaholic.com/novel/i-was-caught-up-in-a-hero-summoning-but-that-world-is-at-peace/
NOVEL: I Was Caught up in a Hero Summoning, but That World Is at Peace
9 volumes and 868 chapters found
? Enter output directory: /Users/crossbx/Desktop/lightnovel-crawler/Lightnovels/www-foxaholic-com/I Was Caught up in a Hero Summoning bu
[...]

PR #1054, opened 2021-08-30
Fix foxaholic body is empty error

watzeedzad:


watzeedzad:
This pull request is fix for #1052

PR #1059, opened 2021-08-31
fix sslv3 handshake error when download image from foxaholic, fix png image save as jpg

watzeedzad:
- Fix getting SSLv3 error when download image from foxaholic using cloudscraper
 - Fix png image saved as jpg image


watzeedzad:
Realize that my code didn't support all image format on foxaholic, I will take a look at it. But right now I will cancel this PR.

PR #1060, opened 2021-08-31
Foxaholic

watzeedzad:
- Fix SSLv3 handshake error while download chapter image on foxaholic (switch from cloudscraper to requests)
 - Add support for .png image in chapter content
 - Force save chapter content image to RGB mode
 - Fix duplicate level one heading in chapter content
 - Fix duplicate source url when --add-source-url is enable in chapter content

### DIFF
--- a/lncrawl/binders/epub.py
+++ b/lncrawl/binders/epub.py
@@ -24,6 +24,8 @@ def make_cover_image(app):
         cover_image.content = image_file.read()
     # end with
     return cover_image
+
+
 # end def
 
 
@@ -75,6 +77,8 @@ def make_intro_page(app, cover_image):
         title='Intro',
         content=intro_html,
     )
+
+
 # end def
 
 
@@ -107,6 +111,8 @@ def make_chapters(book, chapters):
         # end if
     # end for
     book.toc = tuple(toc)
+
+
 # end def
 
 
@@ -116,12 +122,12 @@ def make_chapter_images(book, image_output_path):
     # end if
 
     for filename in os.listdir(image_output_path):
-        if not filename.endswith('.jpg'):
+        if not (filename.endswith('.jpg') or filename.endswith('.png')):
             continue
         # end if
 
         image_item = epub.EpubImage()
-        image_item.media_type = 'image/jpeg'
+        image_item.media_type = map_image_extension_to_type(filename)
         image_item.file_name = 'images/' + filename
         with open(os.path.join(image_output_path, filename), 'rb') as fp:
             image_item.content = fp.read()
@@ -129,6 +135,21 @@ def make_chapter_images(book, image_output_path):
 
         book.add_item(image_item)
     # end for
+
+
+# end def
+
+def map_image_extension_to_type(ext):
+    img_type = ''
+
+    if ext.endswith('.jpg') == 'jpg':
+        img_type = 'image/jpeg'
+    elif ext.endswith('.png') == 'png':
+        img_type = 'image/png'
+
+    return img_type
+
+
 # end def
 
 
@@ -182,6 +203,8 @@ def bind_epub_book(app, chapters, volume=''):
     epub.write_epub(file_path, book, {})
     print('Created: %s.epub' % file_name)
     return file_path
+
+
 # end def
 
 
--- a/lncrawl/binders/epub.py
+++ b/lncrawl/binders/epub.py
@@ -24,8 +24,6 @@ def make_cover_image(app):
         cover_image.content = image_file.read()
     # end with
     return cover_image
-
-
 # end def
 
 
@@ -77,8 +75,6 @@ def make_intro_page(app, cover_image):
         title='Intro',
         content=intro_html,
     )
-
-
 # end def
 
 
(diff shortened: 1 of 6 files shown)


dipu-bd:
There is a reason we are using cloudscraper. It can bypass the cloudflare protected sites. To use requests, we can just override default scraper in the `initialize()` method. No need to modify `crawler.py`. It will break a lot of things. 

Check here for example: https://github.com/dipu-bd/lightnovel-crawler/blob/master/sources/totallytranslations.py

lncrawl/binders/epub.py
```diff
@@ -116,19 +122,34 @@ def make_chapter_images(book, image_output_path):
     # end if
 
     for filename in os.listdir(image_output_path):
-        if not filename.endswith('.jpg'):
+        if not (filename.endswith('.jpg') or filename.endswith('.png')):
```
dipu-bd, line 125: it can never contain png image. this check is entirely unnecessary. check here for how the files are being saved: https://github.com/dipu-bd/lightnovel-crawler/blob/master/lncrawl/core/downloader.py#L159

watzeedzad:
@dipu-bd Didn't know that it is a dump question or not but basically function in specific source can overriden function in `crawler.py` is that right?

watzeedzad:
Regarding your concern I've made change in commit 7a7e65c, please take a look and give me a feedback

dipu-bd:
> @dipu-bd Didn't know that it is a dump question or not but basically function in specific source can overriden function in `crawler.py` is that right?

that is how inheritance works. if you are unfamiliar about it you can read about it in the docs: https://docs.python.org/3.6/tutorial/classes.html#inheritance

dipu-bd (changes requested):
I have a few other questions before approving this.

1. why did you move `download_image` to `crawler.py` from `downloader.py`? I do not want the method to be exposed to all crawlers. 
2. Why are you saving in PNG format instead of JPG? JPG files are much smaller than PNG. loading time in mobile epub readers are faster too.
3. Adding extra new lines before `#end def` is not necessary. seeing all these unnecessary extra new lines just burns my eyes. 👀
4. in .gitignore file `/venv*/` covers `venv/` too.
5. what is `download_image` method doing in foxaholic.py file? it is not being used there anywhere.

watzeedzad:
> I have a few other questions before approving this.
> 
> 1. why did you move `download_image` to `crawler.py` from `downloader.py`? I do not want the method to be exposed to all crawlers.
> 2. Why are you saving in PNG format instead of JPG? JPG files are much smaller than PNG. loading time in mobile epub readers are faster too.
> 3. Adding extra new lines before `#end def` is not necessary. seeing all these unnecessary extra new lines just burns my eyes. 👀
> 4. in .gitignore file `/venv*/` covers `venv/` too.
> 5. what is `download_image` method doing in foxaholic.py file? it is not being used there anywhere.

1.Becase I can overriden the behavior of `download_image` in `foxaholic.py` or other sources.
2.In Foxaholic they also put some PNG format there and I wanted all picture packed into single EPUB. If you don't like that way I think can add a arguments flag to enable or disable PNG format
3.It's because of PyCharm reformating
4.Didn't see it was there, will remove it later
5.So I can override behavior `download_image` when downloading image from Foxaholic to use `requests` instead of `cloudscraper`

dipu-bd:
> 2. In Foxaholic they also put some PNG format there and I wanted all picture packed into single EPUB. If you don't like that way I think can add a arguments flag to enable or disable PNG format.

I guess you might be having a misunderstanding about image format here. No matter what format the original image is, `img.save(f, "JPEG")` will convert it into jpg format before saving it. There is no need to handle png and jpg formats separately, because we are converting everything into jpg.

watzeedzad:
> I guess you might be having a misunderstanding about image format here. No matter what format the original image is, `img.save(f, "JPEG")` will convert it into jpg format before saving it. There is no need to handle png and jpg formats separately, because we are converting everything into jpg.
> 
> Let's not add more complexity to the app by introducing png format and adding an additional input argument needlessly.

Ok you are right, I tired download again and some of PNG image work just fine but some of them are not. So I remove PNG image handling logic in 618388. 

But to fix some PNG image not working I am modify saving image statement to force conversion image to RGB mode.

lncrawl/core/crawler.py
```diff
@@ -115,6 +118,21 @@ def download_chapter_body(self, chapter) -> str:
         pass
     # end def
 
+    def download_image(self, url) -> bytes:
+        assert isinstance(url, str)
+        if len(url) > 1000 or url.startswith('data:'):
+            content = base64.b64decode(url.split('base64,')[-1])
+        else:
+            logger.info('Downloading image: ' + url)
+            response = self.crawler.get_response(url, headers={
```
dipu-bd, line 127: this should be self.get_response

dipu-bd (changes requested):

lncrawl/core/crawler.py
dipu-bd: > But to fix some PNG image not working I am modify saving image statement to force conversion image to RGB mode.

Now it looks good to me. There is a slight issue I mentioned in the comment above

dipu-bd:
> But to fix some PNG image not working I am modify saving image statement to force conversion image to RGB mode.

Now it looks good to me. There is a slight issue I mentioned in the comment above

watzeedzad:
> Now it looks good to me. There is a slight issue I mentioned in the comment above

Fixed in 841942a

dipu-bd (approved):

PR #1072, opened 2021-09-04
Add novelmtl.com source (copy of readwn.com)

watzeedzad:
- Add source novelmtl.com which is copy of readwn.com


watzeedzad:
Sorry for #1071, I created it by mistake

PR #1098, opened 2021-09-20
Fixed a few errors

SirGryphin:
Some sources having problems with ajax on wordpress sites or showing zero chapteters. I've updated the ones I've found.


dipu-bd (approved):

dipu-bd:
Thanks a lot

Issue #1114, opened 2021-09-30
lightnovelpub.com | Error


SirGryphin:
Keep getting `! Error: list index out of range `

**Input**:         https://www.lightnovelpub.com/novel/strongest-eccentric-consort-19072354
**App source**:    PIP
**App version**:   2.28.5
**Your OS**:       Windows

dipu-bd:
Have you  checked the details log using `-lll` options?

SirGryphin:
@dipu-bd 
Didn't even know that was a option lol, I checked logs and found the line of code and commented it out.  I've added to PR #1115 It works fine but I do get a error saying it can't download original cover and downloads a abstract one.

```
2021-10-01 06:31:18,120 [INFO] (lncrawl.core.downloader)
Downloading original cover image...
2021-10-01 06:31:18,121 [INFO] (lncrawl.core.crawler)
Downloading image: https://static.lightnovelpub.com/bookcover/300x400/00961-strongest-eccentric-consort.jpg
2021-10-01 06:31:18,125 [DEBUG] (urllib3.connectionpool)
Starting new HTTPS connection (1): static.lightnovelpub.com:443
2021-10-01 06:31:18,456 [DEBUG] (urllib3.connectionpool)
https://static.lightnovelpub.com:443 "GET /bookcover/300x400/00961-strongest-eccentric-consort.jpg HTTP/1.1" 403 None
2021-10-01 06:31:18,458 [WARNING] (lncrawl.core.downloader)
Failed to download original cover image: https://static.lightnovelpub.com/bookcover/300x400/00961-strongest-eccentric-consort.jpg -> C:\Github\lightnovel-crawler\lncrawl\Lightnovels\www-lightnovelpub-com\Strongest Eccentric Consort\cover.png (403 Client Error: Not allowed for url: https://static.lightnovelpub.com/bookcover/300x400/00961-strongest-eccentric-consort.jpg)
2021-10-01 06:31:18,459 [INFO] (lncrawl.core.downloader)
Downloading fallback cover image...
2021-10-01 06:31:18,460 [INFO] (lncrawl.core.crawler)
Downloading image: https://source.unsplash.com/featured/1200x1550?abstract
```

dipu-bd:
looks like a 403 error is preventing cover to be downloaded. let's skip for now.

PR #1115, opened 2021-09-30
added ltnovel.com

SirGryphin:
Added ltnovel.com to the readwn.py file because the sites are clones of each other, I've tested it and it works. If you want me to make a separate source file let me know. I just thought this would be quicker.


dipu-bd:
Since they are two different sites, they should have two different files. Something could be modified in one site, but not on the other one.

SirGryphin:
Made a separate source file, also included fix for index error I was getting #1114 .

dipu-bd:
Thanks

PR #1118, opened 2021-10-02
Fix foxaholic source get chapter error

watzeedzad:
- As Foxaholic change thier get chapter endpoint from /wp-admin/admin-ajax.php to novel_url/ajax/chapters/

### DIFF
--- a/sources/foxaholic.py
+++ b/sources/foxaholic.py
@@ -60,7 +60,7 @@ class FoxaholicCrawler(Crawler):
         logger.info('Novel id: %s', self.novel_id)
 
         # response = self.submit_form(chapter_list_url, data='action=manga_get_chapters&manga=' + self.novel_id)
-        response = self.submit_form(self.novel_url + 'ajax/chapters/', data='')
+        response = self.submit_form(self.novel_url.strip('/') + '/ajax/chapters/', data='')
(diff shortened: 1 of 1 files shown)


sources/foxaholic.py
```diff
@@ -59,7 +59,8 @@ def read_novel_info(self):
         self.novel_id = soup.select_one('#manga-chapters-holder')['data-id']
         logger.info('Novel id: %s', self.novel_id)
 
-        response = self.submit_form(chapter_list_url, data='action=manga_get_chapters&manga=' + self.novel_id)
+        # response = self.submit_form(chapter_list_url, data='action=manga_get_chapters&manga=' + self.novel_id)
+        response = self.submit_form(self.novel_url + 'ajax/chapters/', data='')
```
dipu-bd, line 63: Since `novel_url` may not always ends with a `/`, it is better to handle such case. A possible solution:

```
response = self.submit_form(self.novel_url.strip('/') + '/ajax/chapters/', data='')
```

dipu-bd (changes requested):

sources/foxaholic.py
watzeedzad: Fixed in 0d84f85

Issue #1195, opened 2021-11-16
Duplicated chapters on RoyalRoad.com


AymaanGillani:
## Describe the bug
The released date link in the royalroad page is getting seen as a chapter on its own and leading to duplicating the same chapter with the title as the release date.

Same thing happens on the bots and on the latest 2.28.11 update

![Issue1](https://user-images.githubusercontent.com/52474239/141993391-6e9bfefb-ce97-45d4-a862-a191225e9408.PNG)
![Issue3](https://user-images.githubusercontent.com/52474239/141993856-179a84dd-5b34-44b9-ad4a-b9830a0f0389.PNG)
![Issue2](https://user-images.githubusercontent.com/52474239/141993328-03b527c1-b1e4-4aa9-b362-cc555657bd83.PNG)
## Let us know

**App source**:    PIP
**App version**:   2.28.5
**Your OS**:       Windows

dipu-bd:
Please provide novel url.

Since this issue is related to a specific source, you should have created a "Source Not Working" issue

aspin:
happening for me too, for example: https://www.royalroad.com/fiction/21220/mother-of-learning

aspin:
Reason for this is that when scanning the table each row contains two elements, the chapter title and the release date, and both elements are being treated as a single chapter (they link to the same result)

PR #1199, opened 2021-11-17
Fix duplicate chapters in Royal Road

aspin:
addresses #1195


junni1337:
you can also add string=True, because it's the only one with a string in the tag

Issue #1223, opened 2021-12-17
https://dragontea.ink/


mesmerlord:
- Language: English
- Supports Searching: Yes
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: Yes
- Has CloudFlare Protection: No

PR #1224, opened 2021-12-17
1223 new source dragon tea

mesmerlord:
@dipu-bd  Connected issue/ticket: https://github.com/dipu-bd/lightnovel-crawler/issues/1223


dipu-bd:
Thanks a lot

dipu-bd:
The `base_url` was not correct. It was missing an `h` at the start. I fixed it anyhow.

PR #1229, opened 2021-12-22
Add a timeout to each search (default 10s)

VexedPanda:
Searches can take minutes, or even stall/fail in ways that prevent the app from continuing / being useful. Most sites seem to (though I haven't gathered real data) return search results within 10s, and the slower ones aren't great candidates for fetching actual chapters from either, so pruning them seems acceptable.

Note that there is one corner case that's worth calling out. The search threads themselves aren't terminated, we just stop waiting for them. As such, the app itself won't exit, and will need two ctrl-cs to exit while the background searches complete running.


dipu-bd:
Thanks for pointing it out. Please do pull request to `dev` branch in the future. I fixed and merged this pull request.

PR #1322, opened 2022-03-31
Add https://wuxia.city support

Joacchim:
Hello,

First of all, thanks for the tool, it's really convenient \o/

I wanted to use it to download one of my ongoing reads from wuxia.city, which was not supported yet, so here's my attempt at adding the support to lightnovel-crawler.

Hopefully, it should fit your requirements, otherwise I'm open to suggestions.

Cheers

NB: I've tried it out to download the novel I was reading, and it seemed to work fine, but I haven't checked everything yet.


dipu-bd:
Thanks. It looks good to me.

PR #1332, opened 2022-04-14
Fix hardcoded novel_id in myboxnovel.py

lotmklein:
Replaced the hard coded manga id with self.novel_id.

The problem before was, that when using MyBoxNovelCrawler, it only got the chapters from https://myboxnovel.com/book/reincarnation-of-the-businesswoman-at-school-boxnovel/ and not from the inputed novel.


dipu-bd:
ops! missed that somehow 😞

lotmklein:
No problem :smile:

dipu-bd:
you are welcome! send me more if you find anything else. there are too many issues but too few people working on them nowadays.

PR #1340, opened 2022-04-20
fixed readwn source

mesmerlord:
Tested with both novels that have more than 1 page of chapters and novels with only one page.

For convenience, you can test with:

More than 1 page: https://www.readwn.com/novel/turns-out-im-a-great-cultivator.html
1 page: https://www.readwn.com/novel/dont-pretend-to-be-the-heroine.html


dipu-bd:
Thanks for the PR. Can you put it on the dev branch?

mesmerlord:
@dipu-bd Done

PR #1342, opened 2022-04-21
Added manga sources

SirGryphin:
I've not posted a source file in a while and the layout and how it works has changed within the last year. I'm not sure what other files I'm meant to edit.

It use to be the README.md to add the source to the "Supported Sites" and that's it. Now there some kind of _index json that I see you edit in latest PR. Could you explain what I need to do when add sources now or is it okay just to PR the single python file.

**UPDATE:** I've add a few manga sites as I'm downloading some for a friend. If there not allowed for support sites please let me know and I will stop pushing them.

### DIFF
--- a/sources/en/c/coffeemanga.py
+++ b/sources/en/c/coffeemanga.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+import logging
+from lncrawl.core.crawler import Crawler
+
+logger = logging.getLogger(__name__)
+search_url = 'https://novelmic.com/?s=%s&post_type=wp-manga'
+chapter_list_url = 'https://novelmic.com/wp-admin/admin-ajax.php'
+
+
+class CoffeeManga(Crawler):
+    has_manga = True
+    base_url = 'https://coffeemanga.com/'
+
+    def search_novel(self, query):
+        query = query.lower().replace(' ', '+')
+        soup = self.get_soup(search_url % query)
+
+        results = []
+        for tab in soup.select('div.c-tabs-item__content'):
+            a = tab.select_one('div.post-title h3 a')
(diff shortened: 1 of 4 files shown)


SirGryphin:
@UltraStudioLTD 
Yeah, I thought they was allowed, there is a feature in light novel crawler that supports them and it works okay.

UltraStudioLTD:
@dipu-bd if this PR gets accepted, you should rename this project (**lightnovel-crawler**) to **manganovel-crawler** or something like that

UltraStudioLTD:
@SirGryphin it just doesn't sit well with current name of project which is novels-only

SirGryphin:
@UltraStudioLTD If @dipu-bd wants to remove it, I see no problem. I was just never told not to do manga sites, I'm 95% sure there are other sources/sites already added with manga. If it's against rules or recommend not to somewhere and I missed it please let me know. I don't want to cause any problems.

dipu-bd:
@SirGryphin With new structure, you do not need to edit any README.md or index.json files. They are autogenerated. Just update the crawler files in `sources` folder. After you commit to dev branch, a auto commit will happen to update README.md, and index.json

dipu-bd:
@UltraStudioLTD  There is no rule saying it can't have manga just because the project name starts with `lightnovel`. Changing project name is a bit of a hassle.

UltraStudioLTD:
> @UltraStudioLTD There is no rule saying it can't have manga just because the project name starts with `lightnovel`. Changing project name is a bit of a hassle.

Alright

SirGryphin:
@dipu-bd I've switch to Dev, sorry about that. Also what your saying is.. I just put my source file e.g `LightNovelSite.py` into `Sources > en > L` and it just automatically does the rest?

Issue #1349, opened 2022-04-27
https://888novel.com


idMysteries:
- Language: en
- Supports Searching: Yes
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection: No

PR #1378, opened 2022-06-04
French sources https://chireads.com and https://lnmtlfr.com/

jere344:
* French sources https://chireads.com/ and https://lnmtlfr.com/ 
* English source https://888novel.com/ ( #1349 )
* Fix typos in sources/_sample.py

***

The sources works when using ```--sources [url]``` but when doing a general search they do not show up in the results, I don't know why.

### DIFF
--- a/sources/en/0/888novel.py
+++ b/sources/en/0/888novel.py
@@ -15,7 +15,7 @@ class Eight88NovelCrawler(Crawler):
     machine_translation = False
 
     def initialize(self) -> None:
-        class Eight88NoovelTextCleaner(TextCleaner):
+        class Eight88NovelTextCleaner(TextCleaner):
             """A copy of TextCleaner that keep double <br> tags"""
 
             def clean_contents(self, div):
@@ -48,7 +48,7 @@ class Eight88NovelCrawler(Crawler):
 
             # end def
 
-        self.cleaner = Eight88NoovelTextCleaner()
+        self.cleaner = Eight88NovelTextCleaner()
 
     def search_novel(self, query):
         query = query.replace(" ", "+")
--- a/sources/en/0/888novel.py
+++ b/sources/en/0/888novel.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+
+import logging
+from lncrawl.core.crawler import Crawler
+from lncrawl.utils.cleaner import TextCleaner
+from bs4 import Comment, Tag
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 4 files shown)


sources/en/0/888novel.py
```diff
@@ -0,0 +1,150 @@
...
+                if self.bad_css:
+                    for bad in div.select(",".join(self.bad_css)):
+                        bad.extract()
+                    # end if
+                # end if
+                for tag in div.find_all(True):
+                    if isinstance(tag, Comment):
+                        tag.extract()  # Remove comments
+
+                    # I want to keep double <br> tags
```
dipu-bd, line 34: Why do you want to keep <br> tags? In some epub readers it makes the gap between two paragraphs too wide for comfort.

sources/fr/lnmtlfr.py
```diff
@@ -0,0 +1,93 @@
...
+            for e in soup.find("div", {"class": "author-content"}).find_all("a")
+        ])
+
+        # Chapter are recuperated from a post request.
+        id = soup.find("input", {"class": "rating-post-id"}).get("value")
+
+        url = "https://lnmtlfr.com/wp-admin/admin-ajax.php"
+        data = f"action=manga_get_chapters&manga={id}"
+        resp = self.submit_form(url, data=data)
+        soup = self.make_soup(resp)
```
dipu-bd, line 59: Updated this file to avoid using requests and BeautifulSoup directly

sources/fr/chireads.py
```diff
@@ -0,0 +1,76 @@
...
+        soup = self.get_soup(self.novel_url)
+        content = soup.find("div", {"class": "conteiner"}).find_all(
+            "div", {"class": "wid"}
+        )
+
+        metadata = content[0]
+        
+        self.novel_cover = self.absolute_url(metadata.find("img").get("src"))
+        
+        self.novel_title = metadata.find("h3", {
```
dipu-bd, line 44: self.cleaner is not necessary. novel title, author, titles in chapters and volumes lists gets formatted later automatically

sources/en/0/888novel.py
jere344: For me it looks like a single block with one ```<br>``` using sumatrapdf on windows or moon+reader on android. 

double br

single br

dipu-bd: ah, I get it. i think i should remove the initialize funtion.

you can use calibre for reading. it has some nice features for formatting ebooks for reading.

PR #1409, opened 2022-06-22
New Source

SirGryphin:
Added newsnovel.net, tested with a few novels and everything seems to be find. I've disabled search because it doesn't work on site either.

### DIFF
--- a/sources/en/n/newsnovel.py
+++ b/sources/en/n/newsnovel.py
@@ -1,0 +1,147 @@
+# -*- coding: utf-8 -*-
+import logging
+import re
+from concurrent import futures
+from lncrawl.core.crawler import Crawler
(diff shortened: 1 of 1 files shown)


SirGryphin:
Updated allnovelfull.py, I've commented out some code because it was causing some chapters to show as empty body.

You can check commented code a bottom of source file under download_chapter_body.

This is what I've noticed between working on non-working chapters.

**Download Chapters Okay:**
okay

**Shows as Empty Body:**
fail

SirGryphin:
Add source for https://www.kitenovel.com, I've tested it with a few novels seems to work okay. Some of the chapters show as badly formatted on some novels but that the site/chapter it self not the crawler.

#580 Source Requests

Issue #1417, opened 2022-07-02
https://reaperscans.com/


SirGryphin:
- Language: English
- Supports Searching: Yes
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: Yes
- Has CloudFlare Protection: Yes (I've tried other crawlers and sometimes get 524)

PR #1422, opened 2022-07-07
added reaperscans

SirGryphin:
I've tested this on a few different novels and it seems to work, there are a few little things that aren't 100%.

This is has paid/premium chapters, I've made it skip these as they would just show as blank anyway. Search seems to work with the couple of novels I tried. It does hang when grabbing chapters but it download fine, the site "reaperscans" can be a bit slow sometimes.

The one problem I couldn't figure out is this, the site has custom CSS for text boxes in chapters (shown in image below)
site_preview

But of course when scraping it just grabs text not all the other stuff from site. So it removes this and it now looks like this:
epub_preview

I don't know how to sort that, if anyone else does please do. I also can't remove but title name from above chapter name. With it being different for every book there was no point in adding it in "bad" words to be extract.

I hope this is still okay to add to crawler maybe I'm being too picky lol.

Request #1417

### DIFF
--- a/sources/en/r/reaperscans.py
+++ b/sources/en/r/reaperscans.py
@@ -105,7 +105,7 @@ class Reaperscans(Crawler):
                 if bad in content.text:
                     content.extract()
 
-        return self.cleaner.extract_contents(contents)
+        return contents
 
     # end def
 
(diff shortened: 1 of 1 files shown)


dipu-bd:
`self.cleaner.extract_contents` does the cleaning of all styles and extra stuffs other than texts. You may skip using it if you want to retain source formatting

dipu-bd:
I think it is better to leave custom css out of epub. It looks bad to some epub readers. e.g.: black text on black background, random font size, spacings between letters and paragraphs etc.

Issue #935, opened 2021-06-21
https://americanfaux.com


zisv:
https://americanfaux.com

SirGryphin:
Created and Added to PR #1426

PR #1426, opened 2022-07-11
New Source

SirGryphin:
Improve reaperscans.py to make it remove all junk text from chapters.

Also added https://americanfaux.com/

I've tested on multiple novels on site seems to work fine, the only little problem is the author code. I've manage to grab author name but it also get translators as well. Other than that everything working fine.

Requested #935


dipu-bd:
I made a small change to reaperscans. It is better to configure the cleaner inside the `initialize` method.

PR #1431, opened 2022-07-16
new source

SirGryphin:
This is a Portuguese site, even though they allows you to translate to other languages, it’s just by using google translate. The actual “main” site is Portuguese.

I wasn’t sure where to put .py file so it went into ES folder. Hope that's okay.

Also sorry about Pull Requests for every source, Sometimes I have time to do some every couple of days other times it's longer (with work) I never know when I will have time so it's just easier get them sent out as soon as there done.

### DIFF
--- a/sources/en/l/librarynovel.py
+++ b/sources/en/l/librarynovel.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+import logging
+from bs4 import Tag
+
+from lncrawl.core.crawler import Crawler
+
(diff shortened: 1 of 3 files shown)


dipu-bd:
If the source content is in Portuguese, it should be in **pt** folder. If chapter contents are in english, you may put it in **en** folder, even if the main site language is Portuguese.

Check the [List of ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for the two letter code of a language for getting the folder names.

Issue #1362, opened 2022-05-25
Fix this source : webnovel.com


thescz:
## Let us know

**Novel URL**:   https://www.webnovel.com/book/my-three-wives-are-beautiful-vampires._20553633305398405)
**App Location**:   /Users/thescz/Library/Python/3.8/bin/lncrawl
**App Version**:   v2.31.0

## Describe this issue

Whenever I try to download any story from Webnovel, it says 0 volumes and 0 chapters found. This occurs with every story on the site. 

![image](https://user-images.githubusercontent.com/106260730/170295612-1b8e15ac-9804-45b4-8163-665e3ad2be6f.png)

RomuloPS:
Example of the new response on their api:

[/get-chapter-list](https://www.webnovel.com/go/pcm/chapter/get-chapter-list?_csrfToken=2e6ea907-f31e-41f3-9162-635e2c166e36&bookId=20553633305398405&pageIndex=0&encryptType=1&_=1653584485676)

Apparently people on the qidian underground discord have figured out the encryption already, if someone wants to talk to them. Their stuff is not open source, so I don't have their code.

budikesuma:
> Example of the new response on their api:
> 
> [/get-chapter-list](https://www.webnovel.com/go/pcm/chapter/get-chapter-list?_csrfToken=2e6ea907-f31e-41f3-9162-635e2c166e36&bookId=20553633305398405&pageIndex=0&encryptType=1&_=1653584485676)
> 
> Apparently people on the qidian underground discord have figured out the encryption already, if someone wants to talk to them. Their stuff is not open source, so I don't have their code.

If you think there are the encryption code on the QI site, you are completely wrong, the real encryption is on the wuxiaworld.com site.  until now various scrapers out there have not managed to solve it.  But for this QI site it only uses the merging/joining method between paragraph ids.  It's very easy to solve.  

I can use manual scripts to combine those id paragraphs into one whole chapter.  Maybe you can give me a QI novel link to try out.

mesmerlord:
@budikesuma I think they're talking about 
![image](https://user-images.githubusercontent.com/76161333/170890024-a53fe58b-e05a-4368-95be-f39f683d88b0.png)

It seems to be only on the recently added chapters

budikesuma:
> chapters

@mesmerlord 

Decoding...?, I don't see the need for that. 
Because I can normally scrape free chapters from webnovel sites.

[...]

PR #1455, opened 2022-08-06
Fix www.webnovel.com parser

idMysteries:
Fix webnovel parser #1362

<100 chapters:
image
445 chapters:
image


idMysteries:
The parser is slow because it downloads all the chapters at once.

idMysteries:
It will be funny if they turn on encryption here.

dipu-bd:
Thanks. I tested it and it is worked well. Nice work!

PR #1459, opened 2022-08-13
fix boxnovelorg

SirGryphin:
Box-novel.org wasn't showing title, I've fixed that.

~~Also when I made this Pull Request there a few extra files show that I've changed which I haven't, they were done a long time ago. I'm not sure why there showing up as new now.~~

So yeah, I think I forgot to submit them last month, so also two new sources for acientheartloss and blackbox-tl.


dipu-bd:
thanks

PR #1466, opened 2022-08-15
WanderingInn

SirGryphin:
Not sure if this site is blocked, I did search rejected list but did not find it. I know they only host one novel but it is quite popular and I need to rip it for myself anyway.


dipu-bd:
I can acccess the site with VPN. Thanks for adding it.

PR #1469, opened 2022-08-16
Madara Crawler

idMysteries:
It is necessary to implement such a template for the parser.

https://github.com/KotatsuApp/kotatsu-parsers/blob/dac06ddd0ba587f5c6333a4610acfde67f16ff33/src/main/kotlin/org/koitharu/kotatsu/parsers/site/madara/MadaraParser.kt


idMysteries:
Multilevel inheritance in python does not work the way I would like. Need to change "core/sources.py "

idMysteries:
``` py
class PianMangaCrawler(MadaraTemplateCrawler):
    def initialize(self):
        self.base_url = 'https://pianmanga.com/'
        self.has_manga = True
# end class

print(PianMangaCrawler.__base__)
```

__base__ is MadaraTemplateCrawler, not 'Crawler'

dipu-bd:
Making common templates for similar crawlers is a very good idea. But to make it work, auto-update module also needs to be changed too. I will make some time to do this update soon. I will leave this PR as open for now.

idMysteries:
I added to the template getting a list of chapters in the new and old ways. I hope this should work.

dipu-bd:
Inheritance in python is a bit tricky because of how the `import` works in python.

idMysteries:
I began to understand python a little better and realized that "self" is a address(pointer, or like & in C) to an instance of a class. This gives me more opportunities to implement my idea.

idMysteries:
Failed to load module ahhh...

idMysteries:
It WORKS!!! AHAHAHHA
@dipu-bd I did it!!

idMysteries:
@dipu-bd 
изображение

idMysteries:
@dipu-bd I tested it on https://mangarockteam.com/

If you like it, then accept these changes.

SirGryphin:
@idMysteries looks good, so by what I understand (very new to python) it this will make it easier makes sources for sites with identical layouts. There are a lot of sites that use Wordpress layout with data id for chapters etc., so this will make it a lot simpler to create a source by just putting in the site url.

The only problem I can see is that some sites have "junk" text in chapters (e.g., Thanks for reading on mysite.com) that you still might have to manually add extra code for if it's possible.

idMysteries:
@SirGryphin yes, you can add extra code here
``` py
class SiteWithTrashCrawler(Crawler):
    has_manga = True
    base_url = [
        'https://sitewithtrash.com/'
    ]

    def initialize(self) -> None:
        madara.initialize(self)
        <EXTRA CLEANER HERE>

```

idMysteries:
Template examples:

https://github.com/manga-download/hakuneko/blob/0985f9bf3ec8f5b585367f50159cccaa649a8ac5/src/web/mjs/connectors/templates/WordPressMadara.mjs

https://github.com/KotatsuApp/kotatsu-parsers/blob/dac06ddd0ba587f5c6333a4610acfde67f16ff33/src/main/kotlin/org/koitharu/kotatsu/parsers/site/madara/MadaraParser.kt

dipu-bd:
Auto update won't work here. Please wait untill I add support for templates to get auto updates. I will work on this with priority.

idMysteries:
Perhaps templates can be implemented better. But as far as I understand, classes in python are much worse than in .NET or C++ (or other langs haha)

In another programming language, it would be possible to implement my idea more beautifully.

dipu-bd:
Implemented the templating system. I modified mangarockteam crawler with this template. Closing this PR since it is solved.

Issue #1453, opened 2022-08-04
Fix this source : Mtlnation


Flwk:
Hello,

I have this error when i try to get a novel on MTLNATION

The novel : https://mtlnation.com/novel/villain-seized-the-assassin-heroine-at-the-start/

```
================================================================================
                          [#] Lightnovel Crawler v2.31.0
                  https://github.com/dipu-bd/lightnovel-crawler
--------------------------------------------------------------------------------

-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://mtlnation.com/novel/villain-seized-the-assassin-heroine-at-the-start/
? Do you want to log in? Yes
? User/Email: *************
? Password: ***************
Retrieving novel info...
https://mtlnation.com/novel/villain-seized-the-assassin-heroine-at-the-start/

 ! Error: list indices must be integers or slices, not str
 <class 'TypeError'>
 File "lncrawl\bots\console\integration.py", line 89, in start
    _download_novel()
  File "lncrawl\bots\console\integration.py", line 78, in _download_novel
    self.app.get_novel_info()
  File "lncrawl\core\app.py", line 140, in get_novel_info
    self.crawler.read_novel_info()
  File "C:\Users\Bilal\.lncrawl\sources\en\m\mtlnation.py", line 55, in read_novel_info
    self.novel_title = data['data']['title']


--------------------------------------------------------------------------------
 -  https://github.com/dipu-bd/lightnovel-crawler/issues
================================================================================
Press ENTER to exit...
```

Thanks a lot !

Flwk:
@dipu-bd Hello,

I tried again,but it doesn't work : 

With login : 

`================================================================================
                          [#] Lightnovel Crawler v2.32.0
                  https://github.com/dipu-bd/lightnovel-crawler
--------------------------------------------------------------------------------

-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://mtlnation.com/novel/is-the-big-villain-crazy-really-made-the-empress-int
o-a-puppet
? Do you want to log in? Yes
? User/Email: *****************
? Password: ***************

 ! Error: 503 Server Error: Service Temporarily Unavailable for url: https://api.mtlnation.com/api/v2/accounts/login
 <class 'requests.exceptions.HTTPError'>
 File "lncrawl\core\crawler.py", line 322, in post_json
    response = self.post_response(url, data, headers)
  File "lncrawl\core\crawler.py", line 262, in post_response
[...]

PR #1479, opened 2022-08-18
Fix https://mtlnation.com/

idMysteries:
Fix https://github.com/dipu-bd/lightnovel-crawler/issues/1453


idMysteries:
Mmm empty body content!!! Fuuu

dipu-bd:
Thanks for the fix

Issue #1465, opened 2022-08-15
Fix this source [lightnovelpub.com]


dastrdly6585:
## Let us know

**Novel URL**:     https://www.lightnovelpub.com/novel/reformation-of-the-deadbeat-noble-31011236, https://www.lightnovelpub.com/novel/the-villainess-lives-twice-11072357
**App Location**:    EXE
**App Version**:   2.31.0
## Describe this issue

Certain novels on lightnovelpub appear to have the chapters incorrectly detected.  When the crawler splits the novel into volumes, each volume is an exact replica of the first volume.

idMysteries:
Doesn't look like a source error @dipu-bd

dastrdly6585:
One common factor between these two novels is that the app claims that both have 400 chapters exactly when they have 358 and 333 chapters. Each volume in the output epub is a copy of the first 100 chapters.  Here's the generated meta.json from one of the two novels (where I've scrubbed the output path):
[meta.txt](https://github.com/dipu-bd/lightnovel-crawler/files/9381848/meta.txt)

idMysteries:
there should be several epub files on the output, it seems.

idMysteries:
There should be several epub files on the output, it seems.

idMysteries:
Oh...the meta file is wrong... Really! OK... I'll look at the code more closely

dastrdly6585:
Just confirmed that this issue appears only in the EXE.  I tested the PIP version and the chapter list returns correctly.

idMysteries:
It looks like you had an old source

idMysteries:
C:\Users\%USERNAME%\.lncrawl maybe it should be deleted

idMysteries:
Have you had this update?
![image](https://user-images.githubusercontent.com/11484976/185622561-19e5865e-40da-4b29-a3d3-9872fbf8fa2f.png)

dastrdly6585:
After reinstalling the PIP version to try to fix something else, the issue is now also popping up in the PIP version.  One thing I'm noticing when using the extra logging is that there are a lot of "301 None" extra lines popping up compared to novels which have chapters correctly being found:

```
https://www.lightnovelpub.com:443 "GET /novel/the-villainess-lives-twice-11072357/chapters/page-2 HTTP/1.1" 301 None
2022-08-19 09:15:25,909 [DEBUG] (urllib3.connectionpool)
[...]

PR #1485, opened 2022-08-19
Fix https://www.lightnovelpub.com/

idMysteries:
Fix https://github.com/dipu-bd/lightnovel-crawler/issues/1465


dipu-bd:
That's quite a nice fix

Issue #1457, opened 2022-08-09
https://sj.uukanshu.com/


yikaii:
- Language: Chinese
- Supports Searching: Yes
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection: Yes/No (no clue)

idMysteries:
Copy https://github.com/dipu-bd/lightnovel-crawler/issues/515
@dipu-bd

idMysteries:
@dipu-bd can you close the issue?

Issue #515, opened 2020-06-27
https://sj.uukanshu.com/


AmanMeenaSulanya:

SirGryphin:
Tried to making source file but I'm just getting gibberish text back, they must of some kind of system in place to block scrapering. I don't know if anyone else will have better luck.

PR #1487, opened 2022-08-19
Add https://sj.uukanshu.com/ 

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/1457
https://github.com/dipu-bd/lightnovel-crawler/issues/515

TODO:
- [x] Search
- [x] Volumes


idMysteries:
Everything seems to be ready. But due to the fact that it is Chinese, you need to have Chinese in the OS to search.
image

dipu-bd:
> But due to the fact that it is Chinese, you need to have Chinese in the OS to search.

You have to enable the utf-8 encoding in the command prompt to see them. It is better to use the Windows Terminal if you are on windows.

dipu-bd:
I see that you left one task unchecked:  

> Better format text maybe

Is this PR ready to merge?

idMysteries:
I didn't notice any problems with format_text
If one of the users finds it, I will fix it later

Issue #1436, opened 2022-07-18
https://bonnovel.com


levinuiiy:
I checked, & none for now

No, it has not been requested

- Language: English
- Supports Searching: Yes
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection: Yes/No, i think its No

mooglestiltzkin:
noticed some chapters were skipped when ripping. but when check the site manually, the chapters were available and working. any ideas why?

http://boxnovel.org/i-have-a-mansion-in-the-post-apocalyptic-world-chapter-652-quantum-computer-research-institute

dipu-bd:
@mooglestiltzkin This is a source request for `bonnovel`. Your issue seems to be about box novel. Create another issue with it.

PR #1488, opened 2022-08-20
Add source https://bonnovel.com/

idMysteries:
Close https://github.com/dipu-bd/lightnovel-crawler/issues/1436


idMysteries:
Some madara sites only work through admin-ajax.php
We have to take this into account

PR #1493, opened 2022-08-21
fixed tipnovel

SirGryphin:
Noticed it wasn't downloading anything. I've managed to get it working now. Download a novel and it working fine now.

**Note:** Clicknovel.net was doing same but when I checked there site, it just gives me `clicknovel.net redirected you too many times.
` error. So I'm not sure what to do with that one.

tipnovel


dipu-bd:
Thanks for the fix.

The issue with clicknovel might be related to the site itself. I am not able to access it with my browser too.

PR #1495, opened 2022-08-22
added https://hotnovelfull.com/

SirGryphin:
Clone site of "readnovelfull.py" everything working but for some reason it won't download cover images. I've tried multiple different ways.

### DIFF
--- a/sources/en/h/hotnovelfull.py
+++ b/sources/en/h/hotnovelfull.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+import logging
+from lncrawl.core.crawler import Crawler
+
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


idMysteries:
The cover image is downloaded normally.

idMysteries (approved):

SirGryphin:
@idMysteries might just be the novel, I'm trying then. but for me the cover was missing. At least it works  😃

idMysteries:
Have you updated the code to version 2.32? Maybe that's the problem? It works for me, I have synchronized the repositories.

idMysteries:
I'm too dumb to understand the difference between our realities. It doesn't work for you, it works for me. Is it magic, science, many-worlds interpretation?
Lately I feel like an idiot. Sorry I'm Russian. Т_Т

PR #1501, opened 2022-08-25
Boxnovel moved to _down?

SirGryphin:
So boxnovel.org has to domains listed one is http and the other is https they show this when trying to download from them.

boxnovel

When I tried searching again to reproduce the error it not showing on search now. So I went to the site(s).
http site:
`Whoops, looks like something went wrong.`
https site:
`Oops! page not found.`

I've also noticed now the https site redirects to https://beautymanga.com/ so it looks like they brought boxnovel domain, Also the novel I was looking for isn't on there site.

I've moved to _down and added to rejected list. Hope this is okay. Also boxnovel/beautymanga got blocked as "bad" by kaspersky not sure if it as false positive because its redirecting.


idMysteries:
http is OK...

изображение

dipu-bd:
boxnovel seems to be okay. beautymanga is maybe a clone of boxnovel. we can add another crawler for it. since http works, no need to move it to _down list.

PR #1526, opened 2022-09-01
fix progress

jere344:
I'm not sure about this, but I believe this cause the ```app.progress``` value to be wrong (in bots, not in the console progress bar).
Both ```finally``` block can sometimes be triggered which increment the progress more than once for a chapter, giving for example 300/170 chapters downloaded.
Removing this seems to fix it for the few novels I tried.


dipu-bd:
you spotted it right. there were two app.progress increment in the same method. i somehow missed that

Issue #1534, opened 2022-09-03
Fix this source [reaperscans.com]


dastrdly6585:
## Let us know

**Novel URL**:     All (e.g. https://reaperscans.com/series/swordmasters-youngest-son/)
**App Location**:    EXE
**App Version**:   2.33.1

## Describe this issue

Crawler is not properly detecting which chapters are locked and producing chapters with empty bodies for those chapters.

PR #1535, opened 2022-09-03
Fix https://reaperscans.com/

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/1534


idMysteries:
@dipu-bd I don't have a premium and I don't know if it will work with a premium. But according to the idea everything should be fine.

Issue #1553, opened 2022-09-07
https://readonlinenovels.com/


graciagraciagracia:
- Language: EN
- Supports Searching: Yes
- Contains Machine Translations: 
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection:

idMysteries:
http://readonlinenovels.com/ 
https://github.com/dipu-bd/lightnovel-crawler/blob/851c6a75e52657935de9c91fe52fcc396808693d/sources/en/r/readonlinenovels.py

PR #1554, opened 2022-09-07
Fix https://readonlinenovels.com/

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/1553


idMysteries:
@dipu-bd The search was disabled. I have tested it and it is not slow. So I decided to turn on the search.

If I made a mistake, delete the search code.

PR #1583, opened 2022-09-15
Global Holy Deletion

idMysteries:
OMG, I did something terrible. But it's beautiful.

### DIFF
--- a/lncrawl/binders/__init__.py
+++ b/lncrawl/binders/__init__.py
@@ -43,7 +43,7 @@ def generate_books(app, data):
     out_formats = app.output_formats
     if not out_formats:
         out_formats = {}
-    # end if
+
     out_formats = {x: out_formats.get(x, False) for x in available_formats}
 
     # Resolve formats to output maintaining dependencies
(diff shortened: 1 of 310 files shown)


idMysteries:
310 changed files @dipu-bd 
AHAHAHAH

lncrawl/binders/__init__.py
```diff
@@ -66,14 +66,14 @@ def generate_books(app, data):
...
+
         except Exception as err:
             logger.exception('Failed to generate "%s": %s' % (fmt, err))
         finally:
             progress += 1
             app.progress = 100 * progress / len(formats_to_generate)
-        # end try
-    # end for
+
+
```
dipu-bd, line 76: it would be better if it did not have these unnecessary newlines

sources/es/domentranslations.py
```diff
@@ -14,7 +14,7 @@ class CclawTranslations(Crawler):
 
     def initialize(self) -> None:
         self.cleaner.blacklist_patterns.update(['CONTENIDO | SIGUIENTE'])
-    # end def
+
```
dipu-bd, line 17: changing sources will cause all users to get a force update. imagine you are running the app for a quick use, and having to wait for 300+ crawlers to fnish downloading. we should not modify all sources for just removing comments

dipu-bd (changes requested):

PR #1585, opened 2022-09-16
source

SirGryphin:
added meownovel.com

Some reason search isn't working for me, its not showing novel. I used this one "The Beginning After The End".


dipu-bd:
I have done some modification to the file. Please check if it is working. I could not, because of their maintenance break.

PR #1596, opened 2022-09-20
myboxnovel fix + sources

SirGryphin:
I noticed in certain chapters on myboxnovel.com there was random numbers, I've added a bit of code to remove them. Also added wuxiaworldsite.com to rejected list as it a clone/same site as wuxiaworld.site and that already on there, hope that's okay.

**Note:** How long does it take to update source file to normal pip/exe I noticed even without a update I will get downloading new sources every so often.

@dipu-bd **QUESTION**: If I come across sites where I can't create source because I keep getting CF requests.exceptions.HTTPError: 403 Client Error: Forbidden for url error should I just add URL to rejected list for now?


SirGryphin:
added 2 manga sites, couldn't get search to work on either one for some reason.

dipu-bd:
> **QUESTION**: If I come across sites where I can't create source because I keep getting CF requests.exceptions.HTTPError: 403 Client Error: Forbidden for url error should I just add URL to rejected list for now?

No, just leave them as it is. If there is an issue open, you can add a comment saying "Could not bypass Cloudflare"

We will work on them later, when we figure out a way to bypass it.

idMysteries:
It's a pity that I don't have the rights to edit the code. I can do a more or less good code review.

dipu-bd:
> How long does it take to update source file to normal pip/exe I noticed even without a update I will get downloading new sources every so often.

After master branch is updated, the sources that are changed gets downloaded. Each new version release is packed with latest sources. New versions are released by manual trigger. I usually do it once every two weeks.

dipu-bd:
> It's a pity that I don't have the rights to edit the code. I can do a more or less good code review.

You can comment on it. Suggest changes to your comment.

PR #1599, opened 2022-09-21
Correctly Minify CSS and Display Double <br>

mymro:
The previous approach to minifying CSS lead to duplicate entries in the CSS and missing closing brackets:
image
image

Setting `<br><br>` to _display none_ lead to some novels not rendering correctly. In my opinion completely removing `<br><br>` is not the way to go. If two `<br>` are too much, we can just replace them by one in code.

### DIFF
--- a/lncrawl/assets/web/style.css
+++ b/lncrawl/assets/web/style.css
@@ -40,8 +40,7 @@ main {
   padding: 0 10px;
 }
 
-p + br,
-br + br {
+p + br{
   display: none;
 }
 
--- a/lncrawl/assets/web/style.py
+++ b/lncrawl/assets/web/style.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import re
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


lncrawl/assets/web/style.css
```diff
@@ -40,8 +40,7 @@ main {
   padding: 0 10px;
 }
 
-p + br,
-br + br {
```
dipu-bd, line 44: why remove this? br followed by a br is not a good news

dipu-bd:
> If two  `<br>`  are too much, we can just replace them by one in code.

The code actually removes the first  `<br>`  follow by a second `<br>` tag. It does not remove both

dipu-bd:
Please send PR to `dev` branch.

dipu-bd:
I actually did this change in the dev branch myself. Closing this PR.

mymro:
@dipu-bd If the double <br> should already be removed, then this code is not working. I have a few books where this is not the case. I will take a look.

Issue #1603, opened 2022-09-22
Fix this source : Unknown


jere344:
## Let us know

**Novel URL**:     the girl who ate a death god
**App Location**:    All
**App Version**:   2.35

## Describe this issue

Some source search is returning a ```novelhard.com``` url, which there is no crawler for.

![image](https://user-images.githubusercontent.com/86294972/191869399-272e6710-59a9-4332-bf8b-4e82bfaf52bb.png)
![image](https://user-images.githubusercontent.com/86294972/191869416-25255c80-d375-449d-8f33-2b7e913be8ea.png)

idMysteries:
Something redirects to novelhard. I am aware of this problem, but I do not know how to find the culprit.

idMysteries:
Okay, I found a way to save the log to a file. @dipu-bd You need to add a setting to save logs to a file. I can't fit everything in the cmd buffer.

idMysteries:
@dipu-bd .... nice log.... The module name was not written.
```
2022-09-23 10:51:04,528 [DEBUG] (lncrawl.core.novel_search)
[...]

PR #1604, opened 2022-09-23
Fix https://novelhard.com/

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/1603

madara

https://noveltrench.com/ -> https://novelhard.com/


dipu-bd:
In cases that we have to remove a source, it is better to add the URL to `_rejected.json`

Issue #1605, opened 2022-09-23
Sources with problems


idMysteries:
I looked at the log file and found a lot of broken sources.

Connection to wuxiaworld.co timed out.
wuxiaworld.co timed out, but m.wuxiaworld.co works!
home_url wuxiaworld.co -> m.wuxiaworld.co?

Connection to novelcrush.com timed out.
The crawler needs to be moved to _down

Connection to readnovelz.net timed out.
The site does not redirect. Someone bought this domain. move to _down

Tests needed:
```
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Service Temporarily Unavailable for url: https://888novel.com/tim-kiem/?title=the&he_liet=yes&status=all

    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Service Temporarily Unavailable for url: https://www.fanfiction.net/search/?keywords=the&type=story&match=title&ready=1&categoryid=202

    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Service Temporarily Unavailable for url: https://www.fictionpress.com/search/?keywords=the&type=story&match=title&ready=1&categoryid=202

    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Service Temporarily Unavailable for url: https://kissmanga.in/?s=the&post_type=wp-manga&author=&artist=&release=

    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 503 Server Error: Service Temporarily Unavailable for url: https://novelgate.net/search/the

```

```
move to _down?
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://clicknovel.net/?s=the&post_type=wp-manga


SEARCH FIX NEEDED:
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://mtlreader.com/search

move to _down?
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://wuxiaworld.io/search.ajax?type=&query=the

```

idMysteries:
```
    raise HTTPError(http_error_msg, response=self)
[...]

PR #1608, opened 2022-09-23
Merge `novelpub.com` and `lightnovelpub.com`

jere344:
They are exactly the same
So #1607 fix this one too

#1605


dipu-bd:
Please do not remove a source file. It may create issues for auto-updates. Also, this does not solves the search URL issue for lightnovelpub.com

jere344:
Ok I didn't know about not removing source. It's just that it was weird to have `novelpub` and `lightnovelpub` in different files while `lightnovelpub` with `lightnovelworld`. The 3 are the same.

And I checked, before it returned nothing, now we get search results, but they are wrong. I will look into it.
image

jere344:
So `lightnovelpub` use a post request to `https://www.lightnovelpub.com/lnsearchlive` with `inputContent=query` but it use cloudflare and cloudscraper does not work so I give up

dipu-bd:
> It's just that it was weird to have `novelpub` and `lightnovelpub` in different files while `lightnovelpub` is with `lightnovelworld`. The 3 are the same

That I know. A new way to make crawlers will be introduced in the next version. Crawlers will be able to use predefined templates. So they will become shorter and it will solve these problems.

dipu-bd:
> So `lightnovelpub` use a post request to `https://www.lightnovelpub.com/lnsearchlive` with `inputContent=query` but it use cloudflare and cloudscraper does not work so I give up

I'll check this out. Thanks for leading it so far

dipu-bd:
Fixed in dev

PR #1611, opened 2022-09-24
Correctly remove multiple <br> and keep more styles

mymro:
The code for removing multiple `<br>` did not work in my case. The cleaner will now also remove  multiple `<br>` if they are separated by white space characters.

Some novels inline styles like bold or italic for effect. Those styles will now also be kept.

Hiding `<br><br>` through CSS leads to issues, because the included CSS will also affect  `<br>text<br>` since the plain text is ignored when applying the selector. I have therefore removed it again. Also, this CSS is not needed, since now there should not be multiple `<br>` after the cleaning step.
image

### DIFF
--- a/lncrawl/assets/epub/__init__.py
+++ b/lncrawl/assets/epub/__init__.py
@@ -1,0 +1,8 @@
+from pathlib import Path
+
+ROOT = Path(__file__).parent
+
+def get_css_style():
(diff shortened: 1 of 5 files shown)


dipu-bd:
Thanks for editing <br> tags. I will simplify code for including font-style or font-weights attributes. So, I reverted the cleaner file here.

PR #1616, opened 2022-09-24
Suppress open folder when Suppress Arugment is set or when inside container

mymro:
Fixes also a bug where opening the folder on linux did not work


dipu-bd:
I forgot to remove comments from the code 😞

Issue #1617, opened 2022-09-24
https://puretl.com/


graciagraciagracia:
- Language: EN
- Supports Searching: No
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection:

graciagraciagracia:
Thank you!

PR #1621, opened 2022-09-24
New source https://puretl.com/

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/1617

@dipu-bd please review this code. This site is difficult to parse and I'm not sure my code is good.

### DIFF
--- a/sources/en/p/puretl.py
+++ b/sources/en/p/puretl.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+
+import logging
+import re
+
+from bs4 import Tag
+from lncrawl.core.crawler import Crawler
+from lncrawl.core.exeptions import LNException
+
+logger = logging.getLogger(__name__)
+
+
+class PeryInfo(Crawler):
+    base_url = ["https://puretl.com/"]
+
+    def initialize(self) -> None:
+        self.cleaner.blacklist_patterns.update(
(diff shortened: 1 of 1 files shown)


sources/en/p/puretl.py
```diff
@@ -0,0 +1,64 @@
...
+                    "title": a.text.strip(),
+                }
+            )
+
+    def download_chapter_body(self, chapter):
+        soup = self.get_soup(chapter["url"])
+        chapter["title"] = soup.select_one(".entry-title").text.strip()
+
+        contents = soup.select_one(".blog-item-content")
+        contents.find("a", string=re.compile("Index")).parent.extract()
```
dipu-bd, line 62: this will throw exception if no such `<a>` element is found

sources/en/p/puretl.py
```diff
@@ -0,0 +1,64 @@
...
+            title_tag["content"]
+            .replace("— Pure Love Translations English Translated Novels", "")
+            .strip()
+        )
+
+        possible_image = soup.select_one("meta[property='og:image']")
+        if isinstance(possible_image, Tag):
+            self.novel_cover = self.absolute_url(possible_image["content"])
+
+        logger.info("Novel cover: %s", self.novel_cover)
```
dipu-bd, line 39: you don't need this logging

PR #1633, opened 2022-09-25
Update scribblehub style

idMysteries:
border -> 2px
padding: 10px
width -> auto
margin: 0 auto
image

### DIFF
--- a/sources/en/s/scribblehub.py
+++ b/sources/en/s/scribblehub.py
@@ -13,7 +13,7 @@ chapter_post_url = "https://www.scribblehub.com/wp-admin/admin-ajax.php"
 
 
 class ScribbleHubCrawler(Crawler):
-    base_url = "https://www.scribblehub.com/"
+    base_url = ["https://www.scribblehub.com/", "https://scribblehub.com/"]
 
     def initialize(self) -> None:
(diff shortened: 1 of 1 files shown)


idMysteries:
@dipu-bd wait...

idMysteries:
You removes author notes?

idMysteries:
I just think that the author's notes should not be deleted, because there is various useful (and not very) information there. For example, explanations to the chapter, pictures, etc.

idMysteries:
As far as I remember, author's notes are not deleted for the webnovel.com site.

idMysteries:
@dipu-bd can we change the div tag to table?
image

idMysteries:
I just don't think that deleting author's notes is the right way.

dipu-bd:
Author notes container spoiler. I removed it because of that. And in most cases when you are binge reading a story, it is sometimes distracting to see anything extra

idMysteries:
If you haven't noticed, the spoiler is implemented there very simply and this program revealed it normally. On the first chapter there was a picture for example.
image

dipu-bd:
Why would a caterpillar be relevant while reading a story though 😛

Issue #1655, opened 2022-09-29
Fix this source : https://allnovel.org, Some paragraphs are missing.


tanreeve:
**Novel URL**:   https://allnovel.org/the-bored-immortal.html
**App Location**:   EXE
**App Version**:  v2.34.0

## Describe this issue

Some paragraphs are missing  in output file (epub).  

## Screenshots

**Original** (page in [allnovel.org](url)) :
![image](https://user-images.githubusercontent.com/114658849/193016056-b573b59f-3c3a-4bb9-af52-d7ee664c9681.png)

**Output** file (epub) : 
![image](https://user-images.githubusercontent.com/114658849/193015305-a15bc9a2-b535-4076-a1a1-b3c43f32ee73.png)

PR #1656, opened 2022-09-29
Fix allnovel

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/1655

### DIFF
--- a/sources/en/a/allnovel.py
+++ b/sources/en/a/allnovel.py
@@ -19,10 +19,11 @@ class AllNovelCrawler(Crawler):
     def initialize(self) -> None:
         self.cleaner.bad_text_regex.update(
             [
-                "If you find any errors ( broken links, non-standard content, etc.. ), "
-                + "Please let us know < report chapter > so we can fix it as soon as possible."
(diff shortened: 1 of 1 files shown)


sources/en/a/allnovel.py
```diff
@@ -59,77 +60,30 @@ def read_novel_info(self):
...
-        content = soup.select_one("div#chapter-content")
+        contents = soup.select_one("div#chapter-content")
 
-        assert isinstance(content, Tag), "No chapter content"
-        for child in content.contents:
-            child.extract()
-            if isinstance(child, Tag) and child.name == "h3":
-                break
+        assert isinstance(contents, Tag), "No chapter content"
+        self.cleaner.extract_contents(contents)
```
dipu-bd, line 87: use `clean_contents` if you are not returning in future changes

idMysteries: Then it is necessary to change other crawlers too.

idMysteries: def clean_contents(self, div) -> None: ??

Issue #644, opened 2020-10-24
https://www.chickengege.org/


lunevale:

idMysteries:
@dipu-bd closed

PR #1660, opened 2022-09-29
New source https://www.chickengege.org/

idMysteries:
https://github.com/dipu-bd/lightnovel-crawler/issues/644

Some chapters are password protected. I don't know how to bypass it.

### DIFF
--- a/sources/en/c/chickengege.py
+++ b/sources/en/c/chickengege.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from bs4 import Tag
+
+from lncrawl.core.crawler import Crawler
+from lncrawl.core.exeptions import LNException
+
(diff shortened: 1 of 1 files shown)


idMysteries:
@dipu-bd bad_css does not work for some reason. "div.class" not removed.

dipu-bd:
You are using `div.go-night-color`. it will remove all div tags with that class name. But I saw other tags containing that class.
image

just pass `.go-night-color`

idMysteries:
Why you don't merge it? I don't use .go-night-color here...
>  it will remove all div tags with that class name. 
No, it doesn't remove the div tag!

idMysteries:
image

dipu-bd:
> No, it doesn't remove the div tag!

interesting.. this should not happen. either it is BeautifulSoup bug or there could be something you are missing

PR #1671, opened 2022-10-02
Correcting small error in chapter.xhtml

zerty:


dipu-bd:
Why do you think it is an error?

zerty:
When opening with calibre, style is not applied
image

dipu-bd:
Why are you directly opening xhtml files? 

In new structure of epub file, I am putting chapters into separate `chapter` folder. The `style.css` is defined one level above the folder. Therefore it is required to use `../style.css`

zerty:
I am oppening the Epub.
For some reason the template is incorrectly applied in the cover which is in the root and not in chapter folder.
The resulting xhtml has the head from the chapter template for unknown reasons
I am guessing it is happening because of some weird behavior in the ebooklib module

dipu-bd:
> The resulting xhtml has the head from the chapter template for unknown reasons
> I am guessing it is happening because of some weird behavior in the ebooklib module

I noticed. This is definitely an issue with ebooklib. I have to do some workaround to fix it.

PR #1718, opened 2022-10-12
updated to new template

SirGryphin:
It had numbers in between the paragraphs, so I've updated the file to use the new template and remove the numbers.

Tested with: https://boxnovel.com/novel/i-got-stronger-by-using-copper-coins-until-i-am-invincible/

**Note:** Using branches for each new source now makes it easier editing if something is wrong and merging if it fine. Hope that's okay.

### DIFF
--- a/sources/en/b/boxnovel.py
+++ b/sources/en/b/boxnovel.py
@@ -1,89 +1,23 @@
 # -*- coding: utf-8 -*-
 import logging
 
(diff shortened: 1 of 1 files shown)


dipu-bd:
> **Note:** Using branches for each new source now makes it easier editing if something is wrong and merging if it fine. Hope that's okay.

That's a good strategy

PR #1725, opened 2022-10-16
Novelupdates.com

SirGryphin:
Not sure if this site is allowed or not, if not just close this and let me know.

Tested: https://www.novelupdates.com/series/sand-sculpture-demon-king-online-farming/

I created a simple source file for novels that link to a blog without a table of content. Like the one listed above. Without using novelupdates to grab  chapters I can't think of another way to get novel from blog.

Update: I know there will be a lot of junk/bad text depending on what blog novelupdates links you to for each novel but it's better than nothing.

### DIFF
--- a/sources/en/n/novelupdatescom.py
+++ b/sources/en/n/novelupdatescom.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+
+import logging
+from bs4 import ResultSet, Tag
+from lncrawl.core.crawler import Crawler
+from lncrawl.models import Chapter
+
+logger = logging.getLogger(__name__)
+
+
+class NovelUpdatesCom(Crawler):
+    base_url = ["https://www.novelupdates.com/", "https://novelupdates.com/"]
+
+    has_manga = False
+    has_mtl = False
+
+    def initialize(self) -> None:
+        self.cleaner.bad_css.update(
+            [
+                "a[href*='ko-fi.com']",
+            ]
+        )
+        self.cleaner.bad_tag_text_pairs.update(
+            {
+                "a": r"ToC",
+            }
+        )
+
+    def read_novel_info(self):
+        logger.debug("Visiting %s", self.novel_url)
+        soup = self.get_soup(self.novel_url)
+
+        possible_title = soup.select_one('meta[property="og:title"]')
+        assert possible_title, "No novel title"
+        self.novel_title = possible_title["content"]
+        logger.info("Novel title: %s", self.novel_title)
+
+        possible_novel_cover = soup.select_one('meta[property="og:image"]')
+        if possible_novel_cover:
+            self.novel_cover = self.absolute_url(possible_novel_cover["content"])
+        logger.info("Novel cover: %s", self.novel_cover)
+
+        possible_author = soup.select_one("div#showauthors a")
+        if isinstance(possible_author, Tag):
+            self.novel_author = possible_author.text.strip()
+        logger.info("Novel author: %s", self.novel_author)
+
+        pagination_link = soup.select(".digg_pagination a:not([rel])")
+        page_count = (
+            int(pagination_link[-1].get_text())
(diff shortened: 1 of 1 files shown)


jere344 (changes requested):
Sometimes novelupdates has several translations for the same chapter, missing chapters or unordered chapters. I think you should try to use the chapter number as id instead of using the order shown on novelupdates

SirGryphin:
@jere344 What something like this?

```python
self.chapters.append( 
     Chapter( 
         id=a["title"].strip() + 1, 
         title=a["title"].strip(), 
         url=self.absolute_url(a["href"]), 
     ) 
 ) 
```

jere344:
@kuwoyuki there is a lot of article reader like app on android that can pull text from any website. I used one for a long time, and was it good enough 70% of the time.
I would really like a novelupdate souce as it has the most content from small fan translation group, even if it does not always work.

But yes the source as is will only work for one website, so it need to be improved.

idMysteries:
What? https://awanderingpotato.com/ for this novel please!

This site is just a redirect to other sites. You should make a separate crawler for each site.

jere344:
@idMysteries making a crawler for each blog with only 1-2 novel isn't worth it, so I think that making an universal crawler for chapter content only using novelupdates metadata can be great.

idMysteries:
> @idMysteries making a crawler for each blog with only 1-2 novel isn't worth it, so I think that making an universal crawler for chapter content only using novelupdates metadata can be great.

The current "lightnovel-crawler" code is designed for the fact that you will use one crawler class for one site or template.

kuwoyuki:
> @idMysteries making a crawler for each blog with only 1-2 novel isn't worth it, so I think that making an universal crawler for chapter content only using novelupdates metadata can be great.

It can obviously work but there's a good chance there will be a lot of broken novels because either the blog DOM is just too messed us for us to analyze it (a lot are) or the link is just dead.  
There are libs used for Web Readability mode such as [mozzila's readability](https://github.com/mozilla/readability) which do a somewhat decent job at analyzing the page DOM and finding the node that is most likely to be the article/content, haven't looked whether there's a Python alternative.

jere344:
> It can obviously work but there's a good chance there will be a lot of broken novels because either the blog DOM is just too messed us for us to analyze it (a lot are) or the link is just dead. Also a good way to do this would be to obviously first attempt to use the existing lncrawl's sources if the link matches one and then fallback to the universal. There are libs used for Web Readability mode such as [mozzila's readability](https://github.com/mozilla/readability) which do a somewhat decent job at analyzing the page DOM and finding the node that is most likely to be the article/content, haven't looked whether there's a Python alternative.

Yes that's what I was thinking about. And I think the benefit of having a novelupdates source which has a lot of original content and the best metadata outweigh the issues with broken novels. But like @idMysteries said it is different from every other sources, so we need @dipu-bd 's opinion.

SirGryphin:
I've been working on a template using the new template system for blog sites, I'm hoping to get it working then use it for novelupdates and see how it works. But to be honest I created this and thought I could update it to work with certain sites, The only time I really need to crawl from NU is when the blog sites doesn't have a TOC page as I would otherwise use something like Web2epub as it will just grab everything from TOC on small blog sites, it quicker and easier than writing a source file for each little blog that pops up.

dipu-bd:
Adding support for novelupdates would give a lot of benefits with chapter indexing and searching novels. Since we have now templates, we can make a general template for novelupdates. Like @kuwoyuki suggested, we can use the exisiting crawlers for parsing chapter contents. And as a fallback we can use a DOM extractor like @jere344 suggested.

With a general novelupdates template it will become much easier to support new sources. We can use the `lookup` bot to make new crawlers very fast. In this case, we need the novelupdates link corresponding to a novel link. I think this can be done using the search of novelupdates.

dipu-bd:
Closing this PR for #1754

PR #1736, opened 2022-10-23
add tigertranslations.py as new english source

camp00000:
As the title states, adds new english source tigertranslations.org

It's a very inconsistent website which makes it a bit difficult to implement all cases but I think I've covered them.

Basically novels can have up to two pages per chapter and the content is just thrown into one big div which has a few additional things that I tried to get rid of (like the next page links for example).

There's a few comments here and there where 

Tested with  https://tigertranslations.org/jack-of-all-trades-7/ and https://tigertranslations.org/makai-hongi/ , seems to work as intended.


dipu-bd:
Please commit to `dev` branch

camp00000:
> Please send Pull Request to `dev` branch

Misread the instructions and screwed up the branch by rebasing dev into it via a PR… I will fix it once I have access to the CLI again as I’ll need to force push so as to not have 100 unnecessary commits in my branch.

Sorry will open a new PR after I’ve done that (prob in a week or two).

PR #1772, opened 2022-11-14
New source Re-library.com

zerty:


sources/en/r/relibrary.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def parse_chapter_item(self, tag: Tag, id: int) -> Chapter:
+        return Chapter(
+            id=id,
+            title=tag.text.strip(),
+            url=self.absolute_url(tag["href"]),
+        )
+
+    def select_chapter_body(self, soup: BeautifulSoup) -> Tag:
+        sleep(5)
```
dipu-bd, line 55: Using sleep is not the solution. You have to reduce the job count.

sources/en/r/relibrary.py
```diff
@@ -0,0 +1,58 @@
...
+    def initialize(self) -> None:
+        self.cleaner.bad_css.update(
+            [
+                ".nextPageLink",
+                ".prevPageLink",
+                ".su-button",
+                "a[href*='re-library.com']",
+            ]
+        )
+        self.cleaner.bad_tag_text_pairs.update({"h2": "References"})
```
dipu-bd, line 25: To reduce job count: `self.init_executor(<your job count here>)`

dipu-bd (changes requested):

sources/en/r/relibrary.py
zerty: Thanks, I did not find the way to do it

zerty: Even with 1 worker, the rate detection system is still firing..., is there anyway to force the maximum rate of item/s?

dipu-bd: It is difficult to determine a proper rate since we don't know exactly how their rate detection works. I do not want to sacrifice the download speed over the chance to get IP blocked. People using proxy will not face this issue. And there is also Google Colab.

dipu-bd (approved):

PR #1784, opened 2022-11-21
Fix filtering by url

mymro:
I am not sure, what the old code tried to do, but it did not work. I changed it to what I believe is the correct implementation.

### DIFF
--- a/lncrawl/core/crawler.py
+++ b/lncrawl/core/crawler.py
@@ -106,7 +106,7 @@ class Crawler(Scraper):
         """Return the index of chapter by given url or 0"""
         url = self.absolute_url(url)
         for chapter in self.chapters:
-            if chapter.url.rsplit("/") == url:
+            if chapter.url == url:
                 return chapter.id
         return 0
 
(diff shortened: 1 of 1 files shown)


lncrawl/core/crawler.py
```diff
@@ -106,7 +106,7 @@ def index_of_chapter(self, url: str) -> int:
         """Return the index of chapter by given url or 0"""
         url = self.absolute_url(url)
         for chapter in self.chapters:
-            if chapter.url.rsplit("/") == url:
```
idMysteries, line 109: maybe rstrip? @dipu-bd

mymro: Might be it

lncrawl/core/crawler.py
```diff
@@ -106,7 +106,7 @@ def index_of_chapter(self, url: str) -> int:
         """Return the index of chapter by given url or 0"""
         url = self.absolute_url(url)
         for chapter in self.chapters:
-            if chapter.url == url:
+            if chapter.url.rstrip("/") == url.rstrip("/"):
```
idMysteries, line 109: ```if chapter.url.rstrip("/") == url:```

bcuz ```url = self.absolute_url(url)```

absolute_url returns url without right "/"

mymro: Changed it

dipu-bd:
Thanks for finding it out

PR #1857, opened 2023-01-18
fix bonnovel cover

jere344:
`src` is a placeholder. Use `data-src` instead.

ps : messed up the commit name and idk how to edit it.


dipu-bd:
There is an edit button right side of the title. You can change from there

PR #1869, opened 2023-01-28
Added language, synopsis and tags support

jere344:
Add support for 3 new types of metadata.
language is set by default with the folder name of the crawler file.

You can test it with this novel `https://chireads.com/category/translatedtales/world-defying-dan-god/` for example as chireads support theses new metadata and it has synopsis and tag.

(summary -> synosis)


lncrawl/binders/calibre.py
```diff
@@ -56,6 +56,12 @@ def epub_to_calibre(app, epub_file, out_fmt):
         file_name_without_ext,
         "--authors",
         app.crawler.novel_author,
+        '--synopsis', 
+        app.crawler.synopsis,
+        '--language', 
+        app.crawler.language,
+        '--novel-tags',
+        app.crawler.novel_tags,
```
dipu-bd, line 64: The calibre does not have these options. Check here for the available options: https://manual.calibre-ebook.com/generated/en/ebook-convert.html

dipu-bd: only the `--language` option is available

lncrawl/binders/epub.py
```diff
@@ -74,6 +80,9 @@ def bind_epub_book(
             <h3>{novel_author}</h3>
         </div>
         <img class="cover" src="{COVER_IMAGE_NAME}">
+        <div class="synopsis">
```
dipu-bd, line 83: instead of putting the synopsis with the cover page, it would be better to make a separate page for it

dipu-bd (changes requested):
- [x] Fix calibre options
- [x] Put synopsis in separate page

lncrawl/binders/calibre.py
jere344: The website linked describe `--comments` as "Set the e-book description" So I guess we should use that for synopsis. And the good options is `--tags` instead of `--novel-tags`

dipu-bd (changes requested):
Can you fix the linting issues: https://github.com/dipu-bd/lightnovel-crawler/actions/runs/4034948833/jobs/6936533942

You can use flake8 for checking available issues.

```
flake8 -v --count --show-source --statistics
```

Issue #1863, opened 2023-01-24
mangastic: domain change, no chapter found


JoshOm21:
Yes

## Let us know

**Novel URL**:     https://mangastic.in/
**App Location**:    Discord | Google Collab
**App Version**:   

## Describe this issue

 The source change. From 
https://mangastic.me to https://mangastic.in
 Even if i changed the site to (.me) still FAILED TO GET NOVEL INFO, NO CHAPTER FOUND.

PR #1875, opened 2023-01-30
Fix 'jaomix' and 'mangastic' source

### LINKED ISSUE #1867: Fix this source (https://jaomix.ru/)
## Let us know

**Novel URL**:     https://jaomix.ru/ (all novels)
**App Location**:    EXE
**App Version**:   3.2.2>

## Describe this issue

I got an error on all novels from https://jaomix.ru/ .
I used to use another program for downloading novels from there but it is no longer supported. Steal I saw that the user created an issue that tags' names were changed.

image

watzeedzad:
Fix #1867 and #1863

### DIFF
--- a/sources/ru/jaomix.py
+++ b/sources/ru/jaomix.py
@@ -34,40 +34,9 @@ class JaomixCrawler(Crawler):
         if img_src:
             self.novel_cover = self.absolute_url(img_src["src"])
 
-        self.novel_id = soup.select_one("div.like-but")["id"]
-
-        pages_soup = self.make_soup(
-            self.submit_form(
-                ajax_url,
-                data={
-                    "action": "toc",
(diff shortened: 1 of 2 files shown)


sources/ru/jaomix.py
```diff
@@ -78,10 +47,58 @@ def parse_chapters(self, page_soup):
                     "id": chap_id,
                     "volume": vol_id,
                     "title": a.text.strip(),
-                    "url": self.home_url.rstrip("/") + a["href"],
+                    "url": self.absolute_url(a['href']),
                 }
             )
 
+        # self.novel_id = soup.select_one("div.like-but")["id"]
```
dipu-bd, line 54: you don't have to keep the previous code in comments. these may never be used in future.

PR #1877, opened 2023-02-02
freewebnovel

SirGryphin:
add more bad tags to remove junk text.


sources/en/f/freewebnovel.py
```diff
@@ -21,16 +21,26 @@ def initialize(self) -> None:
                 "p": [
                     r"freewebnovel\.com",
                     r"Updates by Freewebnovel\. com",
-                    r"” Search Freewebnovel\.com\. on google”\.",
+                    r"\” Search Freewebnovel\.com\. on google\”\.",
```
dipu-bd, line 24: the regex is wrong here. you do not need backslash before the quote. `\”`

please fix all the similar issues with the regex strings here.

dipu-bd (changes requested):

sources/en/f/freewebnovel.py
SirGryphin: changed, hope it's okay now.

dipu-bd:
There were still some issues, I took the liberty to fix them.

PR #1890, opened 2023-02-13
Update downloader.py

zimorok:
fix the error (_on Windows machine, with Virtual Environment. Not sure if Linux has the same problem_)

`! Error: Expecting value: line 1 column 1 (char 0) `

this happen when the JSON file exist but empty, no data inside with 0 size

to reproduce the error
- get any novel (e.g: https://novelhi.com/s/Traffords-Trading-Club)
- download all chapter until finish
- go to download folder>JSON, open any JSON file and delete the data inside then save
- download the same URL to the same Folder
- select to download All chapters
- error above will come out and stop all progress

### DIFF
--- a/lncrawl/core/downloader.py
+++ b/lncrawl/core/downloader.py
@@ -77,13 +77,15 @@ def fetch_chapter_body(app):
         )
         if not os.path.exists(file_name):
             continue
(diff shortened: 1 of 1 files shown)


dipu-bd:
I think it would be better if you use try..except here. Also, please send PR to the `dev` branch.

PR #1914, opened 2023-03-16
Fix RoyalRoad crawler

ebausson:
Change in the fiction index page mage retrieving the title and author impossible.


dipu-bd:
There seems to be merge conflict with the `dev` branch. Can you check the fix in #1914 and resolve your merge conflicts if necessary?

ebausson:
Your 'tag' update fix the issue I was pushing for in the first place, so there is no point in merging it unless dev and master stayed apart for a long time, which doesn't seem to be the case.

Better drop it.

PR #1941, opened 2023-04-17
Added french manga source anime-sama

jere344:
This is a weird one, but it have a lot of good content


jere344:
chapters are in the wrong order, I will fix that

jere344:
It was just a stupid mistake

PR #1946, opened 2023-04-25
Fix indowebnovel info and chapters"

yogasw:
i fix issue when get images, titles and chapter lists


sources/id/indowebnovel.py
```diff
@@ -46,12 +46,12 @@ def read_novel_info(self):
                     "id": chap_id,
                     "volume": vol_id,
                     "url": self.absolute_url(a["href"]),
-                    "title": a.text.strip() or ("Chapter %d" % chap_id),
+                    "title": a.get_text(" - ").strip() or ("Chapter %d" % chap_id),
```
dipu-bd, line 49: This can be used: `a.get_text(" - ", strip=True)`

dipu-bd: Please send pull request to `dev` branch

yogasw: I will create a new PR, because I previously took it from the main branch.

yogasw: I created a new PR here
https://github.com/dipu-bd/lightnovel-crawler/pull/1949/files

dipu-bd: thanks

Issue #1965, opened 2023-05-21
Fix this source


amadeus-96:
The web page of http://ranobes.net/ changed its domain to https://ranobes.top/, that is why an error appears when entering the url of a ranobes novel

**Novel URL**:     https://ranobes.top/
**App Location**:    EXE
**App Version**:   3.2.6

![image](https://github.com/dipu-bd/lightnovel-crawler/assets/67280660/c04d2609-8fb6-4173-b18b-0b54c5cdaccf)

https://github.com/dipu-bd/lightnovel-crawler/assets/67280660/c04d2609-8fb6-4173-b18b-0b54c5cdaccf

dipu-bd:
Fixed. The crawler already covered `ranobes.top`, but two bugs made it unusable once it got there:

- no chapter body selector was set, so every chapter downloaded empty;
- the chapter-list pages were assembled with page 1 first, and page 1 holds the *newest* chapters — so a novel came out starting partway through and out of order.

Verified on a 295-chapter novel: chapters now run 1 → 295 in order, and the first and last both download with full text.

Closing.

Issue #1590, opened 2022-09-19
Ranobes


SirGryphin:
**Novel URL**:    https://ranobes.net/novels/1205184-heavenly-dao-rankings-v140217.html
**App Location**:    PIP
**App Version**:   2.34.0

Cloudflare error, I've seen other message with it can't bypass try google clab and  I did and got this.

```
Input Link & Press Enter: https://ranobes.net/novels/1205184-heavenly-dao-rankings-v140217.html
https://ranobes.net/novels/1205184-heavenly-dao-rankings-v140217.html
Chapter Numbers Range: 1 512
1 512

╭╮╱╱╱╱╱╱╭╮╱╭╮╱╱╱╱╱╱╱╱╱╱╱╱╭╮╱╭━━━╮╱╱╱╱╱╱╱╱╱╭╮
┃┃╱╱╱╱╱╱┃┃╭╯╰╮╱╱╱╱╱╱╱╱╱╱╱┃┃╱┃╭━╮┃╱╱╱╱╱╱╱╱╱┃┃
┃┃╱╱╭┳━━┫╰┻╮╭╋━╮╭━━┳╮╭┳━━┫┃╱┃┃╱╰╋━┳━━┳╮╭╮╭┫┃╭━━┳━╮
┃┃╱╭╋┫╭╮┃╭╮┃┃┃╭╮┫╭╮┃╰╯┃┃━┫┃╱┃┃╱╭┫╭┫╭╮┃╰╯╰╯┃┃┃┃━┫╭╯
┃╰━╯┃┃╰╯┃┃┃┃╰┫┃┃┃╰╯┣╮╭┫┃━┫╰╮┃╰━╯┃┃┃╭╮┣╮╭╮╭┫╰┫┃━┫┃
╰━━━┻┻━╮┣╯╰┻━┻╯╰┻━━╯╰╯╰━━┻━╯╰━━━┻╯╰╯╰╯╰╯╰╯╰━┻━━┻╯
╱╱╱╱╱╭━╯┃ v2.34.0
╱╱╱╱╱╰━━╯ 🔗 https://github.com/dipu-bd/lightnovel-crawler

 ❗ Input is suppressed

[...]

PR #1990, opened 2023-07-01
Rewrite for ranobes.top using browser

zerty:
With search
Issues:
#1590 #1965

### DIFF
--- a/sources/en/r/ranobes.py
+++ b/sources/en/r/ranobes.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 import re
-
+import js2py
 from typing import Generator, Union
 
 from bs4 import BeautifulSoup, Tag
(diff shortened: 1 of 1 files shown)


sources/en/r/ranobes.py
```diff
@@ -1,109 +1,104 @@
...
-class RanobeLibCrawler(Crawler):
+digit_regex = re.compile(r"\/(\d+)-")
+
+
+class RanobeLibCrawler(SearchableBrowserTemplate):
     base_url = [
-        "http://ranobes.net/",
-        "https://ranobes.net/",
+        "https://ranobes.top/",
     ]
```
Seven0492, line 24: It only supports https?

Seven0492: Also, don't want to ask much since I'm not the one doing the work, but would it be possible just to convert .net links to .top?

To maintain cross-compatibility with novels people could already have.

Seven0492, line 24: Might be more sensible to leave `ranobes.net` in and include an `http` link for `.top`.

It wouldn't bring any trouble since ranobes.net redirects to ranobes.top, and it could be implemented with a simple `self.absolute_url`, or something similar.

If we don't consider backwards compatibility, people still using `.net` links, without the appropriate knowledge, would be thoroughly confused, for good reason. Or might even believe that `lncrawler` doesn't support `ranobes`, which is an erroneous, but easy conclusion to make.

zerty: Done in https://github.com/dipu-bd/lightnovel-crawler/pull/1990/commits/68f7c130111092eadd3ac6745a9f1ce8a8ad4685

sources/en/r/ranobes.py
```diff
@@ -1,109 +1,125 @@
 # -*- coding: utf-8 -*-
 import logging
+import re
```
Seven0492, line 3: Also, as an educational exercise, it might be interesting to try using [re2](https://github.com/google/re2), instead.

It's google's implementation of a regex [NFA engine](https://swtch.com/~rsc/regexp/regexp1.html), which is an [ultra-performant](https://swtch.com/~rsc/regexp/regexp1.html) [superset](https://github.com/google/re2/wiki/Syntax) of modern regex implementations, that means that you can use any kind off regex you want, and if it happens to break NFA compatibility, it'll just fall back to standard regex.

It doesn't make any noticeable difference from a user perspective in this specific case, since an url can only be so long. But in other projects of yours this could really come in handy when the data handled start getting into the dozens or hundreds of characters. Or simply because the regex is really complicated.

In practice, this is just a matter of:
- Importing the re2 wrapper as re.
```Python
import re2 as re
```
- Raising an import error in case the Python version is incompatible. (lower than version 2.6)
```Python
try:
    import re2 as re
except ImportError:
    raise ImportError('Python 2.6 or higher is needed for some speed optimizations.')
```
- And import re if re2 is unavailable
```Python
try:
    import re2 as re
except ImportError:
    raise ImportError('Python 2.6 or higher is needed for some speed optimizations.')
else:
    import re
```
Also, apparently, you should have this line just after importing re2:
```Python
re.set_fallback_notification(re.FALLBACK_WARNING)
```
Why? I have no idea, I'm not _that_ familiar with Python.

But, I hope this was useful to you and you learned something today.

Seven0492 (approved):
Looks good

Issue #1942, opened 2023-04-21
[www.scribblehub.com] Fix this source


Zovog1977:
Yes

## Let us know

**Novel URL**:     <your novel url or query>  https://www.scribblehub.com/series/683570/i-got-isekaid-well-shyt/
**App Location**:    PIP | EXE | Discord | Telegram  Discord
**App Version**:   x.y.z   Lightnovel Crawler

## Describe this issue

 When the link is put in, there is a messaged named "no chapters"

<img width="1392" alt="Screenshot 2023-04-21 at 22 01 45" src="https://user-images.githubusercontent.com/110282720/233688318-16daaacb-b834-4691-b6a1-f089bb0766c7.png">

Zovog1977:
Is this problem only related to discord?

dipu-bd:
Closing as a duplicate of #1788, which is the oldest open report for `scribblehub.com` and is where this is tracked.

Nothing is lost by this — subscribe to #1788 for updates. If what you hit is actually a different fault on the same site rather than the same one, say so and this will be reopened.

Issue #1896, opened 2023-02-21
Fix https://www.scribblehub.com : no chapters


Thrymst:
## Let us know

:   **https://www.scribblehub.com**  <your novel url or query> 
**Discord**:    PIP | EXE | Discord | Telegram 
**Lightnovel crawler**:   x.y.z  

## Describe this issue

0 chapter from every novel

jimster6543:
Same problem

![image_2023-02-21_111624802](https://user-images.githubusercontent.com/6154621/220400141-de6fc869-acbd-43e7-9d51-545a3b33a566.png)

Pingger:
Issue applies to all "sources", that use Cloudflare. Interestingly a simple curl or wget does not show any issues (even when hammering with ~300 requests per minute).

Some Sources don't even have the cloudflare captcha detection yet. e.g. RoyalRoad:
```
Retrieving novel info...
Exception in thread Thread-1 (read_novel_info):
Traceback (most recent call last):
  File "threading.py", line 1016, in _bootstrap_inner
  File "threading.py", line 953, in run
  File "C:\Users\<username>\AppData\Local\Temp\_MEI98682\sources\en\r\royalroad.py", line 36, in read_novel_info
    self.novel_title = soup.find("h1", {"property": "name"}).text.strip()
AttributeError: 'NoneType' object has no attribute 'text'

 ! Error: No chapters found
<class 'Exception'>
File "lncrawl\bots\console\integration.py", line 107, in start
    raise e
  File "lncrawl\bots\console\integration.py", line 101, in start
    _download_novel()
  File "lncrawl\bots\console\integration.py", line 85, in _download_novel
    self.app.get_novel_info()
  File "lncrawl\core\app.py", line 137, in get_novel_info
    raise Exception("No chapters found")
```

Issue #1788, opened 2022-11-24
[www.scribblehub.com] Detected a Cloudflare version 2 challenge


dipu-bd:
## Let us know

**Novel URL**:  https://www.scribblehub.com/series/252562/a-sinners-chance/
**App Location**:    Any 
**App Version**:  latest

PR #1992, opened 2023-07-02
Rewrite scribblehub.com with browser

zerty:
issues: #1942 #1896 #1788

### DIFF
--- a/sources/en/s/scribblehub.py
+++ b/sources/en/s/scribblehub.py
@@ -58,10 +58,6 @@ class ScribbleHubCrawler(SearchableBrowserTemplate):
             )
         )
         self.browser.wait(".search")
-        for a in self.browser.soup.select(
-            ".fic .search_main_box .search_body .search_title a"
-        ):
-            print(a)
         for elem in self.browser.soup.select(
             ".fic .search_main_box .search_body .search_title a"
         ):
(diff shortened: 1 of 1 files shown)


sources/en/s/scribblehub.py
```diff
@@ -41,92 +51,79 @@ def initialize(self) -> None:
...
-        table td {
-            padding: 10px;
-        }
-        </style>
-        """
-        return body
+    def parse_chapter_list(
+        self, soup: BeautifulSoup
+    ) -> Generator[Union[Chapter, Volume], None, None]:
+        pass
```
dipu-bd, line 122: missing implementation

zerty: done

sources/en/s/scribblehub.py
```diff
@@ -1,22 +1,32 @@
...
 
+digit_regex = re.compile(r"\?toc=(\d+)#content1$")
 
-class ScribbleHubCrawler(Crawler):
+
+class ScribbleHubCrawler(SearchableBrowserTemplate):
     base_url = [
         "https://www.scribblehub.com/",
         "https://scribblehub.com/",
     ]
```
Seven0492, line 25: Same thing here, it'd be nice if it could also support `http`.

kuwoyuki: What's the use case for this? An end-user will never paste an `http` scheme URL on a website that has an `https` redirect scheme.

zerty: let's stay in https

PR #2022, opened 2023-08-03
Create 69shu.py

jere344:
I had to overwrite get_source to make the request because I didn't know how to deal with the encoding. The other Chinese sources works fine from the start tho, I don't know why

### DIFF
--- a/sources/zh/69shu.py
+++ b/sources/zh/69shu.py
@@ -1,0 +1,133 @@
+# -*- coding: utf-8 -*-
+import logging
+from bs4 import Tag
+from lncrawl.core.crawler import Crawler
+import urllib.parse
+
+import requests
+from bs4 import BeautifulSoup
+
+headers = {
+    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:101.0) Gecko/20100101 Firefox/101.0",
+    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
+    "Accept-Language": "en-US,en;q=0.5",
+    "Content-Type": "application/x-www-form-urlencoded",
+    "Origin": "https://www.69shu.com",
+    "DNT": "1",
+    "Alt-Used": "www.69shu.com",
+    "Connection": "keep-alive",
+    "Upgrade-Insecure-Requests": "1",
+    "Sec-Fetch-Dest": "document",
+    "Sec-Fetch-Mode": "navigate",
+    "Sec-Fetch-Site": "same-origin",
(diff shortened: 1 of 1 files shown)


jere344:
The issue with encoding is that I need to set the encoding of the response to `gbk` before it is parsed to a soup.
I don't know if there is already  a way to do that, if not I can just add an `encoding` argument to `Scraper` `get_soup()` then `make_soup` like so :
```py
def make_soup(
        self,
        data: Union[Response, bytes, str],
        parser: Optional[str] = None,
        encoding="utf8"
    ) -> BeautifulSoup:
        if isinstance(data, Response):
            html = data.content.decode(encoding, "ignore")
        elif isinstance(data, bytes):
            html = data.decode(encoding, "ignore")
        elif isinstance(data, str):
            html = str(data)
        else:
            raise LNException("Could not parse response")
        return BeautifulSoup(html, features=parser or self.parser)
```

current :
```py
def make_soup(
        self,
        data: Union[Response, bytes, str],
        parser: Optional[str] = None,
    ) -> BeautifulSoup:
        if isinstance(data, Response):
            html = data.content.decode("utf8", "ignore")
        elif isinstance(data, bytes):
            html = data.decode("utf8", "ignore")
        elif isinstance(data, str):
            html = str(data)
        else:
            raise LNException("Could not parse response")
        return BeautifulSoup(html, features=parser or self.parser)
```

zGadli:
So, I researched this a little bit. I see two solutions. We make a check on the source for its encoding by looking for charset in the <head></head>. In this case we get "gbk" for 69shu.com and then use a gbk decoding option. 

The other solution would be to use a library to detect encoding. for example: chardet or multilingual.

It would look something like tihs:

```
pip install chardet
import chardet

def detect_encoding(byte_data):
    result = chardet.detect(byte_data)
    return result['encoding']

byte_data = b'\xe8\xbf\x99\xe5\xa3\xb0\xe9\x9f\xb3'  # Example bytes data
detected_encoding = detect_encoding(byte_data)
print(detected_encoding)  # Output: utf-8

```

zGadli:
Just tested using chardet in colab. It can't detect "gbk" encoding. Seems like chardet option is the only one.

zGadli:
This code seems to get the job done.

```
import requests
from bs4 import BeautifulSoup

url = 'https://www.69shu.com/txt/C48057.htm'  # Replace with the URL you want to check

response = requests.get(url)
html_content = response.text

soup = BeautifulSoup(html_content, 'html.parser')

meta_tag = soup.find('meta', {'charset': True})

if meta_tag:
    charset = meta_tag['charset']
    print('Charset:', charset)
else:
    print('Charset information not found in meta tag.')

```
Tested in google colab.

Issue #2010, opened 2023-07-20
https://wuxiaworld.eu/


engthiagoacco:
- Language:
- Supports Searching: Yes/No
- Contains Machine Translations: Yes/No
- Contains Manga/Manhua/Manhwa: Yes/No
- Has CloudFlare Protection: Yes/No

PR #2026, opened 2023-08-04
Create wuxiaclick.py

jere344:
`https://wuxiaworld.eu/` redirect to `https://wuxia.click/` now #2010

It get `429 Client Error: Too Many Requests` quickly, is there a way to add delay between requests ?


zGadli:
Are you in the Discord Server for this project @jere344 . If yes, could you please tell what your username is I would like to have a conversation about this project.

dipu-bd:
There is a way to bypass the concurrency and set it to 0

I have merged the code for now. Let's do that change later.

PR #2029, opened 2023-08-05
Added Sleep time because timeout error

Nabeelshar:
All this temple websites are countering the crawlers when you visit alot urls at the same time and error ``Timeout error``, to counter that have added sleep timeout of total of 30 seconds.
Suggestions

-maybe someone can make this better by looking how websites are check the url like cache ,IP etc

### DIFF
--- a/lncrawl/templates/novelmtl.py
+++ b/lncrawl/templates/novelmtl.py
@@ -93,4 +93,7 @@ class NovelMTLTemplate(SearchableSoupTemplate, ChapterOnlySoupTemplate):
         )
 
     def select_chapter_body(self, soup: BeautifulSoup) -> Tag:
-        return soup.select_one(".chapter-content")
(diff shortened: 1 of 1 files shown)


dipu-bd:
This should not work as intended. The concurrency is being handled by [TaskManager](https://github.com/dipu-bd/lightnovel-crawler/blob/master/lncrawl/core/taskman.py).

Nabeelshar:
It is working, I tried to go any website of that temple but it was not working as I visited the website it was showing timeout error.

> This should not work as intended. T

dipu-bd:
You can now use the ratelimiter. Example:

https://github.com/dipu-bd/lightnovel-crawler/blob/78f53565e9e0abedd0f3c514a2cd35b2cebe533a/lncrawl/templates/novelmtl.py#L19

Issue #2020, opened 2023-08-01
Issue with lncrawl: Failure to Capture <hr /> Tag in Generated EPUB File


CrimsonBloodfang:
I would like to report an issue regarding lncrawl, specifically related to the failure of capturing and preserving the `<hr />` or horizontal rule element in the generated EPUB files.

I am using lncrawl on Windows 11 by downloading the executable file from the GitHub releases page. When I use lncrawl to crawl content and generate EPUB files, I have observed that the horizontal rule created by the `<hr />` tag is not successfully copied or retained in the resulting EPUB files. This becomes noticeable when I read the EPUB files using the Moon+ Reader Pro app on Android.

To address this issue, I suggest considering modifications to the behavior of lncrawl. Instead of discarding the `<hr />` tag, it would be highly beneficial if lncrawl could replace it with a visually equivalent symbol, such as a sequence of asterisks `*****` or anything similar, within the EPUB files. By doing so, the visual separation provided by the horizontal rule would be preserved, ensuring compatibility across various EPUB readers, including Moon+ Reader Pro.

Incorporating this change would greatly enhance the usability and consistency of the EPUB files generated by lncrawl, regardless of the reader software used.

jere344:
The code for this is already there but commented out. Maybe it caused issue elsewhere or it was just forgotten. I'm opening a pull request and we'll see what dipu-dp think

arromdee:
Sounds like there needs to be an option, if including it is bad for some sites and omitting it is bad for other sites, and there's no way for a program to know which.

zGadli:
You can try to generate an epub by forking the project and running it locally. Just make sure to switch to the `dev` branch and apply the changes in #2030 .

PR #2030, opened 2023-08-06
hr separation

jere344:
I don't know if it was commented out voluntarily or forgotten. Did it cause issues elsewhere ?

#2020


dipu-bd:
some webpage has too frequent `hr` tags. the content does not look good for them. it was commented out to prevent such cases

Issue #2031, opened 2023-08-07
Fix this source


zGadli:
## Let us know

**Novel URL**:     https://1stkissnovel.org/novel/overlord-ln/
**App Location**:   EXE
**App Version**:  Lightnovel Crawler 3.2.8

## Describe this issue

The source mentioned doesn't seem to be working properly. 

❯ python lncrawl
================================================================================
                          [#] Lightnovel Crawler v3.2.8
                  https://github.com/dipu-bd/lightnovel-crawler
--------------------------------------------------------------------------------

-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://1stkissnovel.org/novel/overlord-ln/
Retrieving novel info...
Exception in thread Thread-1 (read_novel_info):
Traceback (most recent call last):
  File "Python311\Lib\threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "Python311\Lib\threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "lightnovel-crawler\sources\en\1\1stkissnovel.py", line 53, in read_novel_info
    self.novel_cover = self.absolute_url(img_src["data-src"])
                                         ~~~~~~~^^^^^^^^^^^^
  File "Python311\site-packages\bs4\element.py", line 1573, in __getitem__
    return self.attrs[key]
           ~~~~~~~~~~^^^^^
KeyError: 'data-src'

 ! Error: No chapters found 
<class 'Exception'>
File "lncrawl\bots\console\integration.py", line 107, in start
    raise e
  File "lncrawl\bots\console\integration.py", line 101, in start
    _download_novel()
  File "lncrawl\bots\console\integration.py", line 85, in _download_novel
    self.app.get_novel_info()
  File "lncrawl\core\app.py", line 137, in get_novel_info
    raise Exception("No chapters found")

PR #2032, opened 2023-08-07
Solve source 1stkissnovel.org issue

zGadli:
Solved #2031 .


jere344:
`src` is a placeholder and is empty before js kick in, use `data-lazy-src` instead

zGadli:
> `src` is a placeholder and is empty before js kick in, use `data-lazy-src` instead

Alright.

PR #2040, opened 2023-08-10
Use chardet by default to find the encoding

jere344:
See #2022 for context

I tried a few different sources, and it did not seem to cause any issues. 
I still added it as an argument to be able to overwrite it if chardet can't determine encoding.

### DIFF
--- a/lncrawl/core/scraper.py
+++ b/lncrawl/core/scraper.py
@@ -284,7 +284,7 @@ class Scraper(TaskManager, SoupMaker):
         )
         return response.json()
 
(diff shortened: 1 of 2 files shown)


zGadli:
Check if sources: 69shu.com works with the code.

jere344:
When I remove the custom get_soup It works with 69shu.com with no issue

jere344:
Chardet seems to make a lot of mistake in finding the encoding, I'm not sure if it should get merged. For example it messed up the encoding here : `https://chrysanthemumgarden.com/novel-tl/ygbg/ygbg-1/`

dipu-bd:
I merged this to enable passing custom encoding

PR #2042, opened 2023-08-13
Update chrysanthemumgarden.py

jere344:
_~~I didn't know how to pass the data using the default `submit_form`, I will look further into it~~_
**Fixed thanks to Miraris on discord**

**chrysanthemum garden now have**    _~~an unique password for each novel.~~_   **a password that change every week** . **There is no email/username needed so we can put anything, I didn't know how else I could add support for the password.**

### DIFF
--- a/sources/en/c/chrysanthemumgarden.py
+++ b/sources/en/c/chrysanthemumgarden.py
@@ -1,13 +1,35 @@
 # -*- coding: utf-8 -*-
 import logging
 
+from requests import Response
+
 from lncrawl.core.crawler import Crawler
+import requests
 
 logger = logging.getLogger(__name__)
+
+headers = {
+    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:101.0) Gecko/20100101 Firefox/101.0",
+    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
+    "Accept-Language": "en-US,en;q=0.5",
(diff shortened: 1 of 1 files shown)


ghus3rz:
As far as I am ware the password are changed weekly for every novel and not one password per novel. CG also has something that will scramble texts if detected as scrapping.

sources/en/c/chrysanthemumgarden.py
```diff
@@ -44,8 +62,24 @@ def read_novel_info(self):
...
+
+        if soup.select_one("#site-pass"):
+            payload = {
+                "site-pass": self.password,
+                "nonce-site-pass": soup.select_one("#nonce-site-pass")["value"],
+                "_wp_http_referer": urlparse(chapter_url).path,
+            }
+
+            soup = self.make_soup(
+                self.submit_form(url=self.absolute_url(chapter_url), data=payload),
```
kuwoyuki, line 81:
```suggestion
                self.submit_form(url=self.absolute_url(chapter_url), data=payload, multipart=True),
```

kuwoyuki: you can also pass `headers` and set stuff like `referer` and `user-agent` there in case they do check those

jere344: I don't think they check theses here as it works as is but that's good to know

TimeLimit67:
Hello. Newbie one here. Can you also teach me how to bypass the password when creating epubs? I still don't get it. everything is greek to me.

PR #2044, opened 2023-08-15
Create ranobelib.py

needKVAS:
New source

### DIFF
--- a/sources/ru/ranobelib.py
+++ b/sources/ru/ranobelib.py
@@ -30,18 +30,24 @@ class RanobeLibMeCrawler(Crawler):
                 ";\t\n "
             )
             content = json.loads(text)
+            break
 
         self.novel_title = content["manga"]["rusName"]
         logger.info("Novel title: %s", self.novel_title)
 
-        self.novel_cover = self.absolute_url(
-            soup.select_one(".media-sidebar__cover > img:nth-child(1)")["src"]
-        )
+        possible_image = soup.select_one(".media-sidebar__cover > img:nth-child(1)")
+        if possible_image:
+            self.novel_cover = self.absolute_url(possible_image["src"])
         logger.info("Novel cover: %s", self.novel_cover)
 
-        self.novel_author = soup.select_one(
-            "div.media-info-list__item:nth-child(6) > div:nth-child(2) > a:nth-child(1)"
-        ).text.strip()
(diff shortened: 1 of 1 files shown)


zGadli:
Tried to visit the website by copying the base_url. It gives me a 404 error. Please check if this is an issue or not.

needKVAS:
Base_url works fine for me. I don't think it's a source file issue.

zGadli:
@needKVAS It might be because the website can not be accessed from my country. Maybe it will work if i use a VPN.

zGadli:
@needKVAS I tried using proxies and VPN yet I still can't access this website. It shows 404 error. Somebody else should confirm if they can access the source website,

needKVAS:
@zGadli Yes, the website is really not available for most countries: [https://check-host.net/check-report/112e7396k5a2](https://check-host.net/check-report/112e7396k5a2)

zGadli:
@needKVAS Oh! Thanks for the info.

sources/ru/ranobelib.py
```diff
@@ -0,0 +1,94 @@
...
+
+        for script in soup.find_all("script"):
+            json_var = "window.__DATA__ = "
+            text = script.text.strip()
+            if not text or not text.startswith(json_var):
+                continue
+            text = text[len(json_var) : text.find("window._SITE_COLOR_")].strip(
+                ";\t\n "
+            )
+            content = json.loads(text)
```
dipu-bd, line 32: you could use a `break` after this

sources/ru/ranobelib.py
```diff
@@ -0,0 +1,94 @@
...
+        self.novel_title = content["manga"]["rusName"]
+        logger.info("Novel title: %s", self.novel_title)
+
+        self.novel_cover = self.absolute_url(
+            soup.select_one(".media-sidebar__cover > img:nth-child(1)")["src"]
+        )
+        logger.info("Novel cover: %s", self.novel_cover)
+
+        self.novel_author = soup.select_one(
+            "div.media-info-list__item:nth-child(6) > div:nth-child(2) > a:nth-child(1)"
```
dipu-bd, line 43: handle the case where this tag might not exists. it should not throw error the crawler if novel author tag is not found. the author is optional.

needKVAS:
@dipu-bd  Thanks for the review. Fixed.

Issue #2046, opened 2023-08-17
Fix this source novelsemperor.com


DomID00:
They changed their theme, so it can't find any chapters anymore

https://novelsemperor.com/

zGadli:
Thr crawler file for this source isn't properly written. Someone should check it out.

DomID00:
@zerty it seems to have a problem for the chapters page when I download inside the .txt it says "Failed to download chapter body"

zGadli:
@DomID00 What about other formats?

DomID00:
@zGadli I don't know, I only use the .txt, but I solved the problem, I just had to change the "div" `mx-auto.max-w-3xl`

`    def select_chapter_body(self, soup: BeautifulSoup) -> Tag:
        return soup.select_one(
            "#readernovel, #readerarea, .entry-content,.chap-content,.mx-auto.max-w-3xl"
        )`

`    def visit_chapter_page_in_browser(self, chapter: Chapter) -> None:
        self.visit(chapter.url)
        self.browser.wait(
            "#readernovel, #readerarea, .entry-content,.mainholder,.chap-content,.mx-auto.max-w-3xl"
        )`

DomID00:
@zGadli I also noticed that there are novels that the script does not detect, so throws the error of no chapters, but the novel in question has many chapters

for example this novel: https://novelsemperor.com/series/comprehensive-martial-arts-prince-qin-you-practice-martial-arts-while-i-cultivate-immortals

zGadli:
@DomID00 If the error code is 403 then they are refusing because of server request spam. Use `python lncrawl -lll` to find where the error is originating from.

DomID00:
@zGadli here is the error i got :

```
? Enter novel page url or query novel: https://novelsemperor.com/series/comprehensive-martial-arts-prince-qin-you-pract
ice-martial-arts-while-i-cultivate-immortals
01:32:58 [INFO] (lncrawl.bots.console.integration)
Detected URL input
01:32:58 [INFO] (lncrawl.core.sources)
Initializing crawler for: https://novelsemperor.com/ [C:\Users\Byzz\Desktop\lightnovel-crawler-3.2.8\sources\en\n\novelsemperor.py]
Retrieving novel info...
01:32:58 [DEBUG] (lncrawl.core.scraper)
[GET] https://novelsemperor.com/series/comprehensive-martial-arts-prince-qin-you-practice-martial-arts-while-i-cultivate-immortals
[...]

PR #2049, opened 2023-08-19
Update MangaStream Template for Novels Emperor

zerty:
#2046


dipu-bd:
I think it would be better to create another template for Novels Emperror

DomID00:
@zerty it seems to have a problem for the chapters page when I download inside the .txt it says "Failed to download chapter body"

PR #2051, opened 2023-08-20
support single chapter volumes

HeliosLHC:
Handle single chapter volumes. The current code fails when the regex tries to parse a volume title containing only a single chapter.

E.g. the regex works on volume titles such as "1 - 5", but breaks on "5".


dipu-bd:
Please do merge request to `dev`

HeliosLHC:
> Please do merge request to `dev` branch instead of `master`

Thanks, I have updated the target branch

zGadli:
@dipu-bd probably should merge this in the next version release.

Issue #2045, opened 2023-08-16
There is no such driver by url https://chromedriver.storage.googleapis.com/LATEST_RELEASE_116.0.5845


KijinSeijaa:
## Describe the bug

================================================================================
                          [#] Lightnovel Crawler v3.2.8
                  https://github.com/dipu-bd/lightnovel-crawler
--------------------------------------------------------------------------------

-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://www.lightnovelpub.com/novel/shadow-slave-30071448
Retrieving novel info...
TOC: 100%|███████████████████████████████████████████████████████████████████████████| 11/11 [00:04<00:00,  2.63page/s]

[#] Shadow Slave
11 volumes and 1089 chapters found.
- https://www.lightnovelpub.com/novel/shadow-slave-30071448

? Enter output directory: C:\Users\Name\Downloads\Lightnovels\www-lightnovelpub-com\Shadow Slave
? What to do with existing folder? Remove old folder and start fresh
? Which chapters to download? Custom range using index
? Enter start index (1 to 1089): 759
? Enter final index (1 to 1089): 1089
? 331 chapters selected Continue
? Which output formats to create? [epub]
? How many files to generate? Pack everything into a single file
Chapters:   8%|█████████████▊                                                                                                                                                      | 28/331 [00:06<01:06,  4.53item/s]
Chapters:   0%|                                                                                                                                                                             | 0/331 [00:00<?, ?item/s]Failed to get chapter: There is no such driver by url https://chromedriver.storage.googleapis.com/LATEST_RELEASE_116.0.5845
Chapters:   0%|▍                                                                                                                                                                    | 1/331 [00:05<29:40,  5.40s/item]Failed to get chapter: There is no such driver by url https://chromedriver.storage.googleapis.com/LATEST_RELEASE_116.0.5845
Chapters:   1%|▉                                                                                                                                                                    | 2/331 [00:10<29:01,  5.29s/item]Failed to get chapter: There is no such driver by url https://chromedriver.storage.googleapis.com/LATEST_RELEASE_116.0.5845
Chapters:   1%|█▍                                                                                                                                                                   | 3/331 [00:15<28:45,  5.26s/item]Failed to get chapter: There is no such driver by url https://chromedriver.storage.googleapis.com/LATEST_RELEASE_116.0.5845
[...]

Issue #2014, opened 2023-07-23
Bug Chrome 115.0.5790.102


DomID00:
## Let us know
since I updated chrome to Version 115.0.5790.102, problems start to appear

**Novel URL**:   https://www.webnovel.com/book/hp-the-arcane-thief_20892578405832405  <your novel url or query> 
**App Location**:    EXE 
**App Version**:   3.2.8  

## Describe this issue

================================================================================
                          [#] Lightnovel Crawler v3.2.8
                  https://github.com/dipu-bd/lightnovel-crawler
--------------------------------------------------------------------------------

-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://www.webnovel.com/book/marvel-game-maker-system_21463505405348405
Retrieving novel info...
Exception in thread Thread-1 (read_novel_info):
Traceback (most recent call last):
  File "lncrawl\templates\browser\basic.py", line 88, in read_novel_info
  File "C:\Users\Byzz\.lncrawl\sources\en\w\webnovel.py", line 112, in read_novel_info_in_scraper
    raise FallbackToBrowser()
lncrawl.core.exeptions.FallbackToBrowser

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "threading.py", line 1016, in _bootstrap_inner
  File "threading.py", line 953, in run
  File "lncrawl\templates\browser\basic.py", line 95, in read_novel_info
  File "C:\Users\Byzz\.lncrawl\sources\en\w\webnovel.py", line 116, in read_novel_info_in_browser
    self.visit(f"{self.home_url}{path}/catalog")
  File "lncrawl\templates\browser\basic.py", line 65, in visit
  File "lncrawl\core\browser.py", line 155, in visit
  File "lncrawl\core\browser.py", line 62, in _init_browser
  File "lncrawl\webdriver\__init__.py", line 35, in create_new
  File "lncrawl\webdriver\local.py", line 109, in create_local
  File "lncrawl\webdriver\local.py", line 28, in _acquire_chrome_driver_path
  File "webdriver_manager\chrome.py", line 39, in install
  File "webdriver_manager\core\manager.py", line 30, in _get_driver_path
  File "webdriver_manager\drivers\chrome.py", line 40, in get_driver_download_url
  File "webdriver_manager\core\driver.py", line 51, in get_driver_version_to_download
  File "webdriver_manager\drivers\chrome.py", line 62, in get_latest_release_version
[...]

PR #2054, opened 2023-08-21
Fix to work on Chrome116 with undetected-chromedriver 3.5.3

zerty:
Still need tests on remote, but no idea how to do ...
#2014 
#2045


zerty:
#2045

zGadli:
Please change the version number to 3.5.3. As it supports the latest version of chrome.

dipu-bd:
To test on remote, you have to run a selenium grid using docker and pass the url with `--selenium-grid` option. 

`docker compose up chrome` should start a selenium server at http://127.0.0.1:4444

PR #2068, opened 2023-08-27
Fixed Freewebnovel

SirGryphin:
So I've added two new domain names, as freewebnovel and bednovel are alternative domains for their main site, innread.com.

I've added `self.init_executor(ratelimit=2)` because otherwise I get `HTTPError: 429 Client Error: Too Many Requests`. I've tried a higher number than 2, but the error comes back.

The last change is that I noticed with freewebnovel (and other domains) that it would have random text in chapters like "innread.com", but it would be weird unicode text, which means it can't be searched and removed normally. So I used unicodedata and normalized the chapter text, which removed the weird watermarks, but then it removed formatting, so I had to update again to fix that.

I've tested it on a few novels, and it works fine. It's a bit slow scraping because of the rate limit, but it works, and all chapters are formatted properly without any watermarks.

Sorry, the `unicodedata.normalize` code isn't pretty, and if someone can rewrite it, go ahead, but I'm just not good enough.

**Update:** forgot to add " innread.com" to self.cleaner.bad_tag_text_pairs.update once it normalized done that now lol. Also "bednovel.com" just in case.

### DIFF
--- a/sources/en/f/freewebnovel.py
+++ b/sources/en/f/freewebnovel.py
@@ -21,7 +21,7 @@ class FreeWebNovelCrawler(SearchableSoupTemplate, ChapterOnlySoupTemplate):
             {
                 "p": [
                     r"freewebnovel\.com",
-                    r"innread\.com"
+                    r"innread\.com",
                     r"Updates by Freewebnovel\. com",
                     r"” Search Freewebnovel\.com\. on google”\.",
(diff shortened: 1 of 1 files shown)


dipu-bd:
The `unicodedata` is not on the dependency list. Either we have to add this dependency or we have to figure out an alternative way to clean. It is better to add this support to the cleaner.

dipu-bd:
We can add this unicodedata.normalize operation as an option in cleaner

PR #2070, opened 2023-08-29
Update novelsemperor.py

### LINKED ISSUE #2046: Fix this source novelsemperor.com

zGadli:
Fix #2046.

### DIFF
--- a/sources/en/n/novelsemperor.py
+++ b/sources/en/n/novelsemperor.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 import logging
-
+import re
 from lncrawl.templates.mangastream import MangaStreamTemplate
 
 logger = logging.getLogger(__name__)
+search_url = 'https://novelsemperor.com/series?title=%s&type=novels&status='
 
 
 class NovelsEmperorCrawler(MangaStreamTemplate):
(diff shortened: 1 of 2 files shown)


sources/en/n/novelsemperor.py
```diff
@@ -17,3 +18,85 @@ def initialize(self) -> None:
...
+    def read_novel_info(self):
+        logger.debug("Visiting %s", self.novel_url)
+        pagination_num = self.get_soup(self.novel_url)
+        pagination_num = pagination_num.select_one("ul > li:nth-last-child(2)").text
+        pagination_num = int(pagination_num)
+
+        soup = []
+
+        for i in range(1, pagination_num + 1):
+            soup.append(self.get_soup(f'{self.novel_url}?page={i}'))
```
dipu-bd, line 50: this will fetch pages one by one. it can slow down the crawler signigicantly. you can use parallel executors. example:
https://github.com/dipu-bd/lightnovel-crawler/blob/431c3687871e4da99b4326f2774c70dcd7746244/sources/zh/uukanshu.py#L62-L69

dipu-bd:
Check out #2049 

There is already an attempt to fix this issue. Since Novels Emperor has diverged from the MangaStream template, we can either create a new template, use the extends the basic Crawler template

sources/en/n/novelsemperor.py
zGadli: Thanks for the review. I am new to python so thanks for the advice I'll look into the example.

jere344:
I didn't read the conversation before doing it so I didn't know about #2049 sorry
I fixed the issues I saw and opened a pull request on @zGadli fork, everything seems to be working. zGadli#2
Except the search which is broken in the website itself (always show no result for me)

Issue #2074, opened 2023-09-01
Update Chinese source URL from "https://www.69shu.com/" to "https://www.69shuba.com/"


Aeterno8:
**Problem:**
One of the supported sources for Lightnovel Crawler is the Chinese source "https://www.69shu.com/." However, it appears that the source's domain has changed to "https://www.69shuba.com/." While users are currently redirected from the old domain to the new one when visiting "https://www.69shu.com/," this redirection may not be permanent. Additionally, if users provide a direct link to a novel that uses the new domain ("https://www.69shuba.com/"), it will not work with the current configuration.

**Steps to Reproduce:**

    1. Open the Lightnovel Crawler application.
    2. Try to access a novel from the source "https://www.69shuba.com/", type for example "https://www.69shuba.com/book/48238.htm"

**Expected Behavior:**

The application should be able to access novel without any issues.

**Additional Context:**

    - It's uncertain how long the redirection from the old domain to the new one will be maintained.

**Proposed Solution:**

As a programmer interested in contributing to this project, I plan to work on this issue myself. To address this problem, I intend to update the code of "69shu.py" to reflect domain change. This will ensure that users can continue to access novels from this source without interruption.

My app source:    PIP
My app version:   v3.2.10
My OS:   Linux

Aeterno8:
The issue has been resolved in pull request #2075, so I am now closing this matter.

PR #2075, opened 2023-09-02
Updated the source URL for the Chinese content from '69shu' to '69shuba'

Aeterno8:
This pull request addresses the issue #2074

Updated the source URL for the Chinese content from '69shu' to '69shuba', while retaining support for the old domain ('69shu'). I have preserved the original '69shu.py' file and implemented changes in a duplicate file, which has been renamed to '69shuba.py.' This adjustment is necessary because 'https://www.69shu.com/' now redirects to 'https://www.69shuba.com/', and many users may not be aware of this domain change due to the redirection.

The '69shu.py' file should be deleted once they cease redirecting to the new domain, or when a significant majority of new users are aware of the domain change.


dipu-bd:
It is better to support these two domain in one file. For examples:

https://github.com/dipu-bd/lightnovel-crawler/blob/2dae4e9326f26fff287fc1b39c24a9ec35d77411/sources/en/m/mangaweebs.py#L11

Aeterno8:
> It is better to support these two domain in one file. For examples:
> 
> https://github.com/dipu-bd/lightnovel-crawler/blob/2dae4e9326f26fff287fc1b39c24a9ec35d77411/sources/en/m/mangaweebs.py#L11
> 
> https://github.com/dipu-bd/lightnovel-crawler/blob/2dae4e9326f26fff287fc1b39c24a9ec35d77411/sources/en/f/freewebnovel.py#L15

I have fixed the problem. In fact, I had tried to address it earlier, but I now realize that I mistakenly used "base_urls" instead of "base_url" for the array name at that time.

Issue #2078, opened 2023-09-04
Failed to generate PDF at TravisTranslations


Arisa43fps:
## Describe the bug

I tried to export in EPUB and PDF formats, I got an error when generating in PDF format. I don't know if this is an issue with the source or the application itself.I haven't tried it with other sources.
For the Calibre application already installed with version 5.44.0.

? Enter novel page url or query novel: https://travistranslations.com/novel/15181/shinka-no-mi/
Retrieving novel info...

[#] Shinka No Mi
2 volumes and 128 chapters found.
- https://travistranslations.com/novel/15181/shinka-no-mi/

? Enter output directory: C:\Users\Arisa43fps\Lightnovels\travistranslations-com\Shinka No Mi
? Which chapters to download? Everything! (128 chapters)
? 128 chapters selected Continue
? Which output formats to create? done (2 selections)
? How many files to generate? Pack everything into a single file
Chapters: 100%|████████████████████████████████████████████████████████████████████| 128/128 [01:12<00:00,  1.76item/s]
  Images: 100%|████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  2.01item/s]
Created: Shinka No Mi c1-128.epub
Failed to generate "pdf": 'TravisTranslations' object has no attribute 'novel_language'
Traceback (most recent call last):
  File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\binders\__init__.py", line 66, in generate_books
    outputs[fmt] = make_calibres(app, outputs["epub"], fmt)
  File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\binders\calibre.py", line 106, in make_calibres
    out = epub_to_calibre(app, epub, out_fmt)
  File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\binders\calibre.py", line 62, in epub_to_calibre
    app.crawler.novel_language,
AttributeError: 'TravisTranslations' object has no attribute 'novel_language'
* Task completed

## Let us know

**App source**:    PIP 
**App version**:   3.2.10  
**Your OS**:       Windows

Arisa43fps:
@zGadli Still the same, possible issues with the source.

Failed to generate "pdf": 'TravisTranslations' object has no attribute 'novel_language'
Traceback (most recent call last):
  File "lncrawl\binders\__init__.py", line 66, in generate_books
[...]

PR #2081, opened 2023-09-06
Fix for Calibre PDF generation.

zGadli:
#2078


zGadli:
@kuwoyuki Yeah I guess you forgot to refactor the change in calibre.py.

dipu-bd:
Thanks for the fix! I should have been more careful.

XxJibril:
does that mean .pdf generation is fixed not ? i just tried using the lastest version, checked .pdf only output but it keeps giving me .epub outputs

PR #2083, opened 2023-09-07
Syosetu volume error fix

### LINKED ISSUE #2082: Error Unknown volume for chapter xx in syosetu
## Let us know

**Novel URL**:     https://ncode.syosetu.com/n3652gg/ 
**App Location**:    PIP | EXE | Discord  
**App Version**:   3.2.10  

## Describe this issue

I tried to download Japanese novels from the syosetu site and got an error.
I've attached the log below.

```
Retrieving novel info...
C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\bs4\builder\__init__.py:545: XMLParsedAsHTMLWarning: It looks like you're parsing an XML document using an HTML parser. If this really is an HTML document (maybe it's XHTML?), you can ignore or filter this warning. If it's XML, you should know that using an XML parser will be more reliable. To parse this document as XML, make sure you have the lxml package installed, and pass the keyword argument `features="xml"` into the BeautifulSoup constructor.
  warnings.warn(

 ! Error: Unknown volume for chapter 58
<class 'AssertionError'>
File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\bots\console\integration.py", line 85, in _download_novel
    self.app.get_novel_info()
  File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\core\app.py", line 135, in get_novel_info
    format_novel(self.crawler)
  File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\core\novel_info.py", line 71, in format_novel
    __format_chapters(crawler, vol_id_map)
  File "C:\Users\Arisa43fps\AppData\Roaming\Python\Python310\site-packages\lncrawl\core\novel_info.py", line 52, in __format_chapters
    assert vol_index >= 0 and vol_index < len(
```

zGadli:
#2082 fixed. The volume error is fixed but the content of the chapters downloaded is `_Failed to download chapter body_`.  If someone could look into this error it would be much appreciated.


zGadli:
@dipu-bd Could you please check the latest commit c875acb. The fix feels hacky and I don't know why we are looking for a `body` tag in the soup of chapter contents could you shed some light?

jere344:
We need to `.find("body")` because of a feature of BeautifulSoup. When you make a soup it automatically create `<html>` and `<body>` tags.

For exemple if you have this :

```py
html = "<p>Chapter 1</p><p>Chapter 2</p>"
soup = BeautifulSoup(html, "lxml")

print(soup)
```
it will give you
```html
<html>
 <body>
  <p>Chapter 1</p>  
  <p>Chapter 2</p>
 </body>
</html>
```

But here we only want 
```html  
<p>Chapter 1</p>  
<p>Chapter 2</p>
```

zGadli:
In the commit 908ad65 it seems @dipu-bd added the line ` chapter.body = soup.find("body").decode_contents()` at `lncrawl\core\crawler.py` line `132` . 

The problem is that `soup.find("body")` returns `None` because in the case of this source (or maybe other sources) the chapter contents do not have a `<body>` tag (the `soup` is the soup of `chapter.body`). So this throws an exception which is caught 
```
except Exception as e:
                chapter.body = ""
                chapter.success = False
                if isinstance(e, KeyboardInterrupt):
                    break
```
Now the problem here is that this redefines the `chapter.body` to nothing. This basically means because there is no `<body>` in the soup of chapter contents, the content is redefined to an empty string. Please correct me if I got something wrong. Do keep in mind that I am a beginner in python so I might have misread some code.

jere344:
That's weird, there should always be a body because of `make_soup`
```py
soup = self.make_soup(chapter.body)
# [...]
chapter.body = soup.find("body").decode_contents()
```

zGadli:
@dipu-bd This pr is ready for review.

lncrawl/core/crawler.py
```diff
@@ -128,7 +128,8 @@ def extract_chapter_images(self, chapter: Chapter) -> None:
             img.attrs = {"src": "images/" + filename, "alt": filename}
             chapter.images[filename] = full_url
 
-        chapter.body = soup.find("body").decode_contents()
+        if soup.find("body") is not None:
```
dipu-bd, line 131: this will result in empty chapter.body

PR #2139, opened 2023-09-21
novel-bin.net

SirGryphin:
**Note:** _Did do another pull request but closed it because I getting checking error. I thought maybe I used older dev version and forgot to update so I update and create source file again it still fails on something to do with telegram which I haven't touched._

Added source for another version of novelbin sites. Created a file as I needed to use Rate limit because of too many request error. 

Also @dipu-bd I'm not sure if you want to combined certain novelbin sources, I know if there same site you just add alternative domains to the base_url. 

It just it noticed the following sources:
novelbin.py = https://novelbin.com/
novelbin.net.py = https://novelbin.net/
novel-bin.py =https://novel-bin.com/

and my one
novel-bin.net.py = https://novel-bin.net/

I think there all the same site, I'm not 100% sure though could be different, I know novel-bin.net had the novel I need and none of the other did so it could just be using same name and site template, it's also another reason I created separate source.

### DIFF
--- a/sources/en/n/novel-bin.net.py
+++ b/sources/en/n/novel-bin.net.py
@@ -1,0 +1,14 @@
+import logging
+
+from lncrawl.templates.novelfull import NovelFullTemplate
+
+logger = logging.getLogger(__name__)
+
+
+class Novel_Bin_Net(NovelFullTemplate):
+    has_mtl = False
+    has_manga = False
+    base_url = ["https://novel-bin.net/"]
+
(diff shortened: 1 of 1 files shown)


jere344:
There's also 
https://allnovelbin.net/ 
https://novel-bin.org/
https://novelbin.me/

dipu-bd:
If they can be parsed with the same code, they can be combined together. Otherweise we should make separate crawlers.

PR #2150, opened 2023-09-30
Fix PDF exporting with docker

rdonaldwilson:
When attemping to convert/export a PDF with the docker image calibre's ebook-convert complains about missing dependiences. Adding libxcomposite-dev and libxdamage-dev fixes it.


dipu-bd:
Thanks

PR #2193, opened 2023-11-10
Memory optimization

jere344:
While profiling memory to find a memory leak I found out that surprisingly this concatenation was 6% of total memory usage. Replacing it with a .join optimized it to 1.5%.
Closing the image was also a smaller but significant improvement.


dipu-bd:
Nice finding

PR #2155, opened 2023-10-04
Optimize Event Listeners and Refactor Code for Enhanced Efficiency

Universe8888:
Function Hoisting: Moved goToHref function outside of the "keyup" event listener to prevent redundant redeclaration, enhancing performance.

Error Handling: Enhanced error handling within the "scroll" event listener to log errors to the console, aiding in debugging.

Cached Selectors: Optimized selector queries by suggesting caching of reused selectors outside of event listeners to minimize DOM querying overhead, though not reflected in the code.

Debouncing: Introduced a debouncing utility to the "scroll" event listener, reducing the frequency of updates to the reading progress indicator and thereby minimizing computational strain. This debouncing utility employs a 100ms delay, ensuring a balanced responsiveness to user scrolling activity and computational efficiency.

PR #2196, opened 2023-11-12
A better version of PR #2155. 

zGadli:
I think the PR #2155 is kinda dead and the author is not implementing the needed improvements so here we go.


jere344 (approved):
Python part looks good.
Opened a small pr on your branch for full quality cover for whitemoon

PR #2202, opened 2023-11-24
Fix Calibre on macOS

### LINKED ISSUE #2192: Calibre on mac not recognized for converting to PDF etc
I installed calibre from dng as normal
running lncrawl will not use calibre

Install Calibre to generate pdf: https://calibre-ebook.com/download

**App source**:    PIP
**App version**:   3.3.2
**Your OS**:      Mac 14.0

CryZFix:
Fix #2192
Changed path to calibre executable file on macOS

P.S. You may object that the installed Calibre from HomeBrew is also able to work directly with "ebook-convert", since a Symlink is created when it is installed. The solution in this commit, work for both installation options, from  a DMG and HomeBrew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,14 @@ An app to download novels from online sources and generate e-books.
 
 **This application uses _Calibre_ to convert ebooks.** <br>
 **Install it from https://calibre-ebook.com/download** <br>
-Without it, you will only get output in epub, text, and web formats.
+Without it, you will only get output in epub, text, and web formats. <br>
+
+**For macOS, you need to manually add the path to Calibre.** <br>
+Before starting lncrawl, use the command:
+```bash
+$ export PATH="$PATH:/Applications/calibre.app/Contents/MacOS"
+```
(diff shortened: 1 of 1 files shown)


dipu-bd:
Can you not export a path variable to MacOS to make the `ebook-convert` globally accessible. like:

```
export PATH="$PATH:/Applications/calibre.app/Contents/MacOS"
```

CryZFix:
> ```
> export PATH="$PATH:/Applications/calibre.app/Contents/MacOS"
> ```
Yes, it will work, but it's slightly not obvious to macOS users who rarely use the terminal.

CryZFix:
> ```
> export PATH="$PATH:/Applications/calibre.app/Contents/MacOS"
> ```
Да, это будет работать, но это немного не очевидно для пользователей macOS, которые редко используют терминал.

dipu-bd:
What if the user has installed the Calibre in another folder other than `/Applications/calibre.app/Contents/MacOS/ebook-convert`? 

This tool is not for general users anyway. Lets add some documentation in the README on how to add the `ebook-convert` to PATH variable, and be done with it.

Issue #2241, opened 2024-01-25
Novelupdates: Fix this source


greenaar:
Yes

**Novel URL**:     Any on Novelupdates, eg https://www.novelupdates.com/series/yigret
**App Location**:    PIP 
**App Version**:   v3.4.0

## Describe this issue

This source was working earlier today with this version of the application, and just stopped working within the last 8 hours.  Below is a debug output from the example series scrape:

Namespace(log=3, log_file=None, list_sources=False, crawler=[], novel_page='https://www.novelupdates.com/series/yigret', query=None, login=None, output_formats=[], add_source_url=False, single=True, multi=False, output_path=None, filename=None, filename_only=True, force=False, ignore=False, all=True, first=None, last=None, page=None, range=None, volumes=None, chapters=None, proxy_file=None, auto_proxy=False, bot=None, shard_id=0, shard_count=1, selenium_grid=None, suppress=True, ignore_images=False, close_directly=False, extra={})
16:49:04 [DEBUG] (lncrawl.core.sources)
Loading current index data from /root/.lncrawl/sources/_index.json
16:49:04 [DEBUG] (lncrawl.core.sources)
Current index was already downloaded once
16:49:04 [DEBUG] (lncrawl.core.sources)
Saving current index data to /root/.lncrawl/sources/_index.json
16:49:04 [DEBUG] (lncrawl.core.sources)
Saving current index data to /root/.lncrawl/sources/_index.json

➡ Press  Ctrl + C  to exit

16:49:04 [INFO] (lncrawl.core.app)
Initialized App
16:49:04 [INFO] (lncrawl.bots.console.integration)
Detected URL input
16:49:04 [INFO] (lncrawl.core.sources)
Initializing crawler for: https://www.novelupdates.com/ [/data/python/lncrawl/lib/python3.10/site-packages/sources/multi/novelupdates.py]
Retrieving novel info...
16:49:04 [DEBUG] (lncrawl.core.scraper)
[GET] https://www.novelupdates.com/series/yigret
timeout=(7, 301), allow_redirects=True, proxies={}, headers={b'Origin': b'https://www.novelupdates.com', b'Referer': b'https://www.novelupdates.com/', b'User-Agent': b'Mozilla/5.0 (X11; Ubuntu; Linux i686; rv:105.0) Gecko/20100101 Firefox/105.0'}
16:49:04 [DEBUG] (urllib3.connectionpool)
[...]

PR #2242, opened 2024-01-25
Fix novelupdates [In process]

CryZFix:
Changed "action" key in ajax
#2241

### DIFF
--- a/lncrawl/templates/novelupdates.py
+++ b/lncrawl/templates/novelupdates.py
@@ -111,7 +111,7 @@ class NovelupdatesTemplate(SearchableBrowserTemplate, ChapterOnlyBrowserTemplate
         response = self.submit_form(
             "https://www.novelupdates.com/wp-admin/admin-ajax.php",
             data=dict(
-                action="nd_getchapters",
+                action="nd_getchapters_full",
                 mygrr="1",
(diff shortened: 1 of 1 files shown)


greenaar:
Worked, but now fails again. I suspect that means it changed again? How did you determine the new value? (ie so that I know where to look?) Thanks!

CryZFix:
Hurried with the conclusions, the problem is local for the book out of the issue. I'm looking at what can be done.

picnixz:
The issue is that the AJAX request needs special cookies. I had a quick look at what was done when you are logged in to NU and requested the full chapters list. When I was logged off, you can't view the full chapters list. After some trials and errors, I deduced that you need to query the API with a cookie acknowledging that you are connected to NU. 

I suspect that for guests, rendering is done on the server side and that a somewhat fake cookie is being used or the check is by-passed if the request is internal (don't know how it's implemented). 

As such, a solution would be to tell users to create a NU account and we would use the cookies used for that (though thezy need to truste how we use their cookies then). I don't see another way to do it since the URLs to the chapters are only available if you are connected (not all endpoints require you to be connected though).

dipu-bd:
The title is marked as In process. Is it still correct?

PR #2243, opened 2024-01-26
Fix Syosetu and Fanstrans

NilanEkanayake:
Syosetu has added pages, so I added a check to see if the source has multiple pages, and if so, iterate through them and scrape as normal. (Was a bit of a rush job, technically works but probably has issues)

Also fixed fanstrans on the side.

I locally fixed a couple of other sources too, might commit them later once I've checked them over

### DIFF
--- a/sources/en/f/fanstrans.py
+++ b/sources/en/f/fanstrans.py
@@ -38,7 +38,7 @@ class FansTranslations(Crawler):
 
     def initialize(self) -> None:
         self.cleaner.bad_tags.update(["h3"])
-        self.init_executor(4) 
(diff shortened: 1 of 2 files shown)


NilanEkanayake:
I added
`
    def initialize(self) -> None:
        self.init_executor(2)
`
But it might not be necessary for most. I was downloading and few extremely long (~600 chapter) series, and Syosetu rate limited me.

dipu-bd:
@NilanEkanayake I have made some quick changes in syosetu. can you please confirm if the changes are okay?

PR #2245, opened 2024-01-28
add wtrlab multilingual source

### LINKED ISSUE #1437: https://wtr-lab.com/en
I checked & not request yet, 

- Language: English
- Supports Searching: Yes
- Contains Machine Translations: Yes
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection: No

### LINKED ISSUE #2222: https://wtr-lab.com/en
- Language: EN
- Supports Searching: Yes
- Contains Machine Translations: Yes
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection: No

https://wtr-lab.com/en

camp00000:
Closes #2222 
Closes #1437 

Tested search and download manually, seems to work just fine.

Basically the same as #2239 as they use the same framework (nextjs, angular or something like that, not sure...)

I am looking into maybe adding a new template so these could be added more easily but there are some strong differences in URLs, JSON Keys and URLs in general for both websites so I'm not sure how viable that actually is.

A lot can also be read out of the json itself as it has the templates for URIs in some cases I believe but that's not necessarily a given at all times so it's pretty opaque at the moment.

### DIFF
--- a/sources/multi/wtrlab.py
+++ b/sources/multi/wtrlab.py
@@ -20,11 +20,14 @@ class WtrLab(Crawler):
         But luckily all necessary data is stored in a consistent JSON that's always in the same script tag
         Essentially the same framework as webfic though with some other keys, urls, etc.
     """
+
     base_url = ["https://wtr-lab.com/"]
     has_manga = False
     has_mtl = True
+    host = ""
 
-    host = base_url[0][:-1]  # remove trailing /
+    def initialize(self) -> None:
+        self.host = self.home_url[:-1] if self.home_url.endswith("/") else self.home_url
 
     def read_novel_info(self):
         soup = self.get_soup(self.novel_url)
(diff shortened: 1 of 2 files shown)


sources/multi/wtrlab.py
```diff
@@ -0,0 +1,133 @@
...
+        The whole site obfuscates classes via Angular or some type of minifier
+        so the only constant HTML comes from display text & HTML IDs
+        But luckily all necessary data is stored in a consistent JSON that's always in the same script tag
+        Essentially the same framework as webfic though with some other keys, urls, etc.
+    """
+    base_url = ["https://wtr-lab.com/"]
+    has_manga = False
+    has_mtl = True
+
+    host = base_url[0][:-1]  # remove trailing /
```
dipu-bd, line 27: please assign this inside the `initialize` method

```
def initialize(self):
    self.host = self.base_url[0][:-1]
```

camp00000: Thanks for the input, moved that part into initialize() & used home_url instead, definitely nicer.
Also added a `return []` at the end of the search func so it returns an empty list if no results are found, looks cleaner that way.


f90208a

PR #2248, opened 2024-01-31
Change in Termux installation manual

pmosko:
- there is no webp package
- added python-lxml and python-grpcio

Termux do not have webp package and I am not sure if it even is needed. 
Installing lxml and grpcio fails on compilation. There are workarounds, but time consuming and unreliable. Fortunatelly thermux have them now as packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ $ docker run -v ~/Lightnovels:/home/appuser/app/Lightnovels -it lncrawl
 - Install [Termux](https://play.google.com/store/apps/details?id=com.termux) from playstore.
 - Open the app and run these commands one by one:
   - `pkg upgrade`
-  - `pkg install python libxml2 libxslt libjpeg-turbo webp rust`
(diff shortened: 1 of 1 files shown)


laundmo:
When running `pip install lightnovel-crawler` it still tries to compile lxml manually, even though the termux package is installed. this errors and fails the entire lightnovel-crawl installation. any idea how to reolve?

FunMan1995:
My work around for termux [proof](https://photos.app.goo.gl/rH9eVuEmRCQ4tdeT8)

pkg upgrade -y && termux-change-repo && pkg install x11-repo -y && termux-setup-storage && pkg upgrade -y && pkg install tsu binutils libxml2-python libxslt libjpeg-turbo rust python-lxml python-pillow python-libsass python-xlib python-grpcio python-apt python-cryptography python-tkinter -y && CFLAGS="-Wno-error=incompatible-function-pointer-types" pip install -U setuptools lightnovel-crawler

cd ~/storage/downloads
lncrawl

PR #2251, opened 2024-02-02
Update royalroad.py

needKVAS:
Removing invisible anti-plagiarism paragraphs


pmosko:
I have mixed feelings about this. On one hand I locally made similar change, on another multiple RR novels where hit by scammers and published on Amazon

PR #2266, opened 2024-02-10
QoL: Miscellanous Fixes

camp00000:
1. Fixes a quite annoying when debugging, exception at the end of a successful download when it logs the zipping of the output.
2. Replaces all logger.warn (deprecated function) calls with logger.warning.

If the 2nd commit is not wanted as it changes quite a few files I can revert it and set the branch back to just fixing the logging call fix mentioned in the first point above.

Also if you have any other suggestions for feasible fixes let me know and/or tag me in the respective issues.


dipu-bd:
thanks

PR #2277, opened 2024-02-22
fix: reaperscans

### LINKED ISSUE #1702: Fix this source: reaperscans.com
## Let us know

**Novel URL**:     All (e.g., https://reaperscans.com/novels/4251-leveling-with-the-gods) 
**App Location**:    EXE 
**App Version**:   3.0.1  

## Describe this issue

Reaperscans.com has completely redone its website, so the crawler now detects no chapters.

kuwoyuki:
some cancer laravel framework and ratelimits
closes #1702

### DIFF
--- a/sources/en/r/reaperscans.py
+++ b/sources/en/r/reaperscans.py
@@ -101,8 +101,7 @@ class Reaperscans(Crawler):
         ]
         self.resolve_futures(toc_futures, desc="TOC", unit="page")
         for f in toc_futures:
-            dom = f.result()
-            chapters.extend(self.get_chapters_from_doc(dom))
+            chapters.extend(self.get_chapters_from_doc(f.result()))
 
         self.insert_chapters(chapters)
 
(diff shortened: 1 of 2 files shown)


sources/en/r/reaperscans.py
```diff
@@ -22,72 +24,99 @@ def initialize(self):
...
-            self.chapters.append(
-                {
-                    "id": chap_id,
-                    "volume": vol_id,
-                    "title": a.text.strip(),
-                    "url": self.absolute_url(a["href"]),
-                }
-            )
+
+        chaps = self.get_chapters_from_doc(container)
```
dipu-bd, line 98: why are we inserting? we could do `append` after reversing the `chaps` array.

kuwoyuki:
done but `resolve_futures` doesn't seem to work properly with `LOG_LEVEL=DEBUG`

dipu-bd:
For `LOG_LEVEL=DEBUG`, it just does not show the progress bar. 

I had trouble scraping from reaperscans before as well. one day it works, the other day requests fail frequently. we can test using different ratelimits for this. e.g. a value of 0.5 might fare better.

I will merge this commit as it is now. If you have found a good ratelimit for reaperscans please make another PR

Issue #2270, opened 2024-02-15
https://www.uukanshu.net/#gsc.tab=0


itsyahma:
- Language:
- Supports Searching: Yes/No
- Contains Machine Translations: Yes/No
- Contains Manga/Manhua/Manhwa: Yes/No
- Has CloudFlare Protection: Yes/No

camp00000:
I've already added this in #2264  -> please close the issue.

This is currently not released yet in a new version, 
if you're technically adept you can use the dev branch which already has the source supported otherwise you'll have to wait for the next official update which will contain this.

itsyahma:
how do i you use the dev branch?

itsyahma:
because i used the dev install but still the same

![image](https://github.com/dipu-bd/lightnovel-crawler/assets/55912690/af4bfaee-000a-4e36-be85-67f3c9339a42)

but when i try to use it it comes out like this
![image](https://github.com/dipu-bd/lightnovel-crawler/assets/55912690/42211521-cc64-441a-ae47-5f3f4f4ce4bb)

camp00000:
Hmm that's very strange, I just tried installing the dev according to the readme, same as you did (just in a virtualenv) and it worked without issue.
![image](https://github.com/dipu-bd/lightnovel-crawler/assets/76262772/7c52e9ec-b53e-4734-95d1-76936bac6940)

camp00000:
@itsyahma can you try running lncrawl with debug output and upload the log here (or to a pastebin for example if it's too long)?
To get debug output you need to add the argument `-lll` (three small L's) when you run lncrawl from your powershell.
If it fills your console too much you can also let it log into a file directly with `--log-file=path_to_file.txt`


Alternatively you can also try running `lncrawl --list-sources` and see if the website's even in there, though it should be...

itsyahma:
so basically i just reinstalled lncrawl and this is what is happening now

camp00000:
This was actually an unhandeled case where the structure of the page was slighty different compared to what I found on when testing generally.

I fixed it in the PR #2279 -> if you're very tech savvy you can use my branch directly to test, otherwise I guess wait for it to be merged into dev and then update your dev version of lncrawl.

I tested the URL you posted in the screenshot and it now works with the adjustments along with previously functional URLs so that seems to work fine, also downloading the first 10 chaps worked fine.

PR #2279, opened 2024-02-23
fix uukanshu possible failure on synopsis

camp00000:
Found on #2270 

This source suffers from rate-limiting after a certain point (only X amount of requests per hour/day allowed, etc) with cloudflare or similar blocking any further requests after a certain point for some time -> is there any way to force proxies or make a popup appear on all log levels for the user to let them know of this limitation that can/should be bypass with proxies if it occurs?

### DIFF
--- a/sources/zh/uukanshu.py
+++ b/sources/zh/uukanshu.py
@@ -38,7 +38,12 @@ class UukanshuOnline(Crawler):
 
         self.novel_title = meta.select_one("h1 > a").text
         self.novel_author = meta.select_one("h2 > a").text
-        self.novel_synopsis = meta.select_one("h3 > p").text
+        synopsis = meta.select_one("h3")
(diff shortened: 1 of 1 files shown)


dipu-bd:
There is configuration for ratelimiting here: `self.init_executor(ratelimit=0.5)` 

Check https://github.com/dipu-bd/lightnovel-crawler/blob/dev/lncrawl/core/taskman.py#L56

dipu-bd:
I think making a message appearing to the user everytime they try to download a novel can become annyoing to some. However, we can add a section in the REAMDE to describe usage of proxies and ratelimiting.

camp00000:
Thanks for the info on the rate_limiting, I realized I may have confused this source's rate-limiting with that of another since I can't find the comments/issues I had in mind right now, but regardless I'll keep this in mind going forward and maybe will create a PR updating the README at some point in the future.

PR #2293, opened 2024-03-03
Core adjustments (typing, misc fixes, replacing os.path with pathlib.Path)

camp00000:
**This builds up on #2292 (first commit is the same).**

It's mostly just some general cleanup in typing, theoretical race conditions and removing os.path usage wherever possible within the core files.
I was considering replacing the leftover `import os` statements in core with `from os import environ, getenv` and so on as that's mostly where it's still used but I've left those as is for now. 

There is no significant performance gain / loss, just some "going with the times" as I find os.path to be a bit antiquated, especially since this project is moving in the direction of no longer supporting older python versions anyways.

I don't mind this being dropped if you find it too risky or it collides with some other major overhaul that's in the works, just let me know with a comment and close the PR in that case.

### DIFF
--- a/lncrawl/core/cleaner.py
+++ b/lncrawl/core/cleaner.py
@@ -4,7 +4,7 @@ import sys
 import unicodedata
 from typing import AnyStr, Dict, List, Set, Union
 
-from bs4 import Comment, Tag, PageElement
+from bs4 import Comment, Tag
 
 LINE_SEP = "<br>"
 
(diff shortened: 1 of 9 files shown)


lncrawl/core/cleaner.py
```diff
@@ -210,14 +203,14 @@ def clean_text(self, text) -> str:
         )
         return text
 
-    def extract_on_duplicate_sibling(self, tag: Tag):
+    def extract_on_duplicate_sibling(self, tag: TAG_LIKE):
```
dipu-bd, line 206: This is not actually necessary. These functions will always get called with `tag: Tag` type. Check the `clean_contents` method conditions:

```
  elif not isinstance(tag, Tag):
                continue  # Skip elements that are not a Tag
```

Issue #2306, opened 2024-03-18
webtoons.com


nixos-s:
webtoons.com is mainly image based comics

- Language:
- Supports Searching: Yes
- Contains Manga/Manhua/Manhwa: Yes
- Has CloudFlare Protection: No

zGadli:
Close the issue the source will be added in the next update.

PR #2307, opened 2024-03-20
Add source webtoon

zGadli:
Add source webtoon #2306


dipu-bd:
@zGadli I can see lots of `Revert` commit in this merge request. Doing this kind of revert commits can be the reason why your header changes disappeared. Please close this and open a new PR again with a clean commit.

PR #2325, opened 2024-04-06
Bug fix

### LINKED ISSUE #2324: ! error decoding to str need a bytes-like object nonetype found
I am getting the error in the title when trying to search a novel, I'm using python.

zGadli:
Fixes #2324

### DIFF
--- a/lncrawl/core/novel_search.py
+++ b/lncrawl/core/novel_search.py
@@ -6,6 +6,7 @@ import os
 from concurrent import futures
 from typing import Dict, List
 
+from bs4 import Tag
 from slugify import slugify
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


lncrawl/core/novel_search.py
```diff
@@ -43,9 +44,19 @@ def _perform_search(app, link, bar):
 def _combine_results(results: List[SearchResult]) -> List[CombinedSearchResult]:
     combined: Dict[str, List[SearchResult]] = {}
     for item in results:
+
+        if item.title is None:
+            logger.warn(f'Title is type None in {item}')
+            continue
+        elif isinstance(item.title, Tag):
+            logger.warn(f'Title is type Tag in {item}')
+            item.title = item.title.get_text()
```
dipu-bd, line 53: this check should not be necessary, as it is always expected to receive a string here. we have to fix the crawlers instead to ensure passing a string title

zGadli: The program completely halts when there is None or Tag type variable passed into the function slugify. I think it's a better alternative to skip using `continue` if we encounter a None or Tag so that the program is less prone to crash but if you want to me to remove the if statements reply to this comment and I will.

zGadli:
@dipu-bd If you don't want the if statements checking for None or Tag type variables passed into this function you can close this pull request and merge the new pull request.

lncrawl/core/novel_search.py
dipu-bd: Sorry for the delayed response. The root cause of the crash is passing invalid type of data to the method. It can be prevented by introducing a type checking. I will merge this commit, and modify the code in a successive commit.

Issue #2326, opened 2024-04-07
cloudflare protection issues


Ateile:
so recently, ive been tryin to download a novel but i cant, more like i couldnt because the cloudflare keeps making it error so it had me like RAHHHH
hope this gets attention

dipu-bd:
can't do anything against cloudflare. we all feel the same pain too

zGadli:
@dipu-bd maybe check this out: [nodriver](https://github.com/ultrafunkamsterdam/nodriver) this package claims to be able to bypass Cloudflare and claims to be "the official successor of the [Undetected-Chromedriver](https://github.com/ultrafunkamsterdam/undetected-chromedriver/) python package".

dipu-bd:
@zGadli This one seems great. Have to test it. Thanks for bringing it to my attention.

Af-50:
@zGadli how do i use nodrive with crawler to get the chapters?

zGadli:
@Af-50 Checkout nodriver documentation. You can get the html of the whole page and parse it using bs4 .

PR #2341, opened 2024-04-15
Temporary fix for foxaholic (turnstile)

watzeedzad:
- Implement seleniumbase to foxaholic which able to bypass turnstile

Notes:
- Still unable to figure out how to persist browser window when downloading chapter contents
- Current implementation for every chapters or images it will open 1 browser window then close. May need to update `BasicBrowserTemplate` class to implement `seleniumbase `
- But at least it work for now XD

### DIFF
--- a/lncrawl/core/novel_search.py
+++ b/lncrawl/core/novel_search.py
@@ -85,7 +85,7 @@ def search_novels(app):
         for item in f.result() or []:
             if not (item and item.title):
                 continue
-            key = slugify(item.title)
+            key = slugify(str(item.title))
             if len(key) <= 2:
                 continue
             combined.setdefault(key, [])


zGadli:
using nodriver is probably a better idea. check #2326

watzeedzad:
Tested with `https://www.foxaholic.com/novel/i-was-caught-up-in-a-hero-summoning-but-that-world-is-at-peace` 

Able to crawl all of tthe chapters and images but the time the duration is quite long (8-9 hours).


@zGadli For nodriver I not too sure about it as I saw that we need to put async in front for the method and did not sure whether or not it will impact overall functionality of the application or not.

watzeedzad:
@zGadli Also if you insist on go with driverless route. There is another lib call `selenium-driverless` as well.

But I will not implement it is I already achive my goal on made epub of `I Was Caught up in a Hero Summoning, but That World Is at Peace` for myself already.

Issue #2344, opened 2024-04-17
Fix this source wtr-lab.com


DomID00:
## Let us know

**Novel URL**:    https://wtr-lab.com/en/serie-4328/the-low-level-trainer-of-the-elf-world
**App Location**:    EXE 
**App Version**:   3.5.1

## Describe this issue

-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://wtr-lab.com/en/serie-6767/from-the-man-behind-the-football-king-to-the-f
ootball-king
Retrieving novel info...
Exception in thread Thread-1 (read_novel_info):
Traceback (most recent call last):
  File "threading.py", line 1016, in _bootstrap_inner
  File "threading.py", line 953, in run
  File "C:\Users\Byzz\.lncrawl\sources\multi\wtrlab.py", line 62, in read_novel_info
    url = f"{self.host}{lang}/serie-{serie_id}/{novel_slug}/chapter-{chapter['slug']}"
KeyError: 'slug'

 ! Error: No chapters found
<class 'Exception'>
File "lncrawl\bots\console\integration.py", line 107, in start
    raise e
  File "lncrawl\bots\console\integration.py", line 101, in start
    _download_novel()
  File "lncrawl\bots\console\integration.py", line 85, in _download_novel
    self.app.get_novel_info()
  File "lncrawl\core\app.py", line 137, in get_novel_info
    raise Exception("No chapters found")

Press ENTER to exit...

DomID00:
@zGadli hi, can you check it pls ↑

zGadli:
@DomID00 Sure

zGadli:
@DomID00 Issue fixed. Close Issue.

DomID00:
@zGadli I made the modification, but it shows me the same error

zGadli:
@DomID00 is your dev branch up to date?

DomID00:
@zGadli yes it works, THANK YOU for solving the problem

PR #2346, opened 2024-04-18
Update wtrlab.py

zGadli:
#2344


jere344:
Since the dev branch started using ` | `, python 3.8 and 3.9 are not supported anymore so the tests are failing.
(it's python 3.10 feature). I don't know if it was voluntary to drop support for older version, if it is the tests should be removed.

PR #2353, opened 2024-04-21
Add continuous navigation when arrow key is held down

zGadli:
When you hold down the left or right arrow key when using the web version of the novel it will continuously traverse through the chapters.

### DIFF
--- a/lncrawl/assets/web/script.js
+++ b/lncrawl/assets/web/script.js
@@ -16,7 +16,7 @@ window.addEventListener("keydown", function (evt) {
         goToHref(document.querySelector("a.prev-button"));
         navigationInterval = setInterval(() => {
           goToHref(document.querySelector("a.prev-button"));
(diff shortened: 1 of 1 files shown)


lncrawl/assets/web/script.js
```diff
@@ -1,24 +1,42 @@
...
+  switch (evt.key) {
+    case "ArrowLeft":
+      keyPressTimer = setTimeout(() => {
         goToHref(document.querySelector("a.prev-button"));
-        break;
-      case "ArrowRight":
+        navigationInterval = setInterval(() => {
+          goToHref(document.querySelector("a.prev-button"));
+        }, 50);
+      }, 10);
```
dipu-bd, line 20: we should add more delay to start auto navigation. like: 500ms

zGadli: ok

PR #2374, opened 2024-05-12
Scribblehub: fix author, add tags and synopsis 

Mause:


needKVAS:
1. Now you have `def read_novel_info(self) -> BeautifulSoup:` **and** `def read_novel_info_in_scraper(self) -> None:` which actually returns `BeautifulSoup`. I think you should revert the first one.

2. I think you should do something like:

```
class GeneralBrowserTemplate(BasicBrowserTemplate, GeneralSoupTemplate):
    # ...

    def read_novel_info_in_scraper(self) -> None:
        # ...
        self.parse_additional_info(soup)
        # ...

    def read_novel_info_in_browser(self) -> None:
        # ...
        self.parse_additional_info(self.browser.soup)
        # ...

    def read_novel_info_in_browser(self, soup: BeautifulSoup) -> None:
        pass
        
    # ...
```
to support browser. This option can also be used in other classes that inherit `GeneralBrowserTemplate`.
I find it counterintuitive that
In `BasicBrowserTemplate`:  `def read_novel_info_in_scraper(self) -> None:`
In `GeneralBrowserTemplate`:  `def read_novel_info_in_scraper(self) -> BeautifulSoup:`
In `ScribbleHubCrawler`: `def read_novel_info_in_scraper(self) -> None:`

dipu-bd:
two separate `parse_genres` and `parse_summary` was added for additional info

PR #2409, opened 2024-07-03
Fix navigation.

zGadli:
When the left or right arrow key is held down you will quickly navigate to the next or previous chapters. I added this feature in a previous pr ( #2353 ) but it's not working right now.


dipu-bd:
Does this PR solve the issue?

zGadli:
@dipu-bd yeah. You can test it if you want.

PR #2455, opened 2024-09-05
Add searching for piaotian

Zokhoi:
There is anti-spam for searching on piaotian, a 20 second limit between two searches. It shouldn't affect much though.

### DIFF
--- a/sources/zh/piaotian.py
+++ b/sources/zh/piaotian.py
@@ -2,6 +2,28 @@
 import logging
 
 from lncrawl.core.crawler import Crawler
+import urllib.parse
+
+headers = {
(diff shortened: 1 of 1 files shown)


sources/zh/piaotian.py
```diff
@@ -2,6 +2,28 @@
 import logging
 
 from lncrawl.core.crawler import Crawler
+import urllib.parse
+
+headers = {
```
dipu-bd, line 7: does it not work without these headers? the cloudscraper automatically set headers using rotating user agents. if we have these headers, that feature will not be used.

dipu-bd (changes requested):

sources/zh/piaotian.py
Zokhoi: Is removing the `User-Agent` header alone enough?

Zokhoi: Is removing `User-Agent` enough for the feature to work?

PR #2516, opened 2024-12-06
Update centralnovel.py

Magos-Technicus:
this way the code will limit the request rate to 3
and thus, make the software work on this website because the website blocks you when the number of requests per second becomes more than 3


dipu-bd:
Please change the target branch to `dev`

Magos-Technicus:
> Please change the target branch to `dev`

Done, thanks for your attention

dipu-bd:
Thanks for the PR. Indentation was missing for the initialize function, I fixed it by the way.

Issue #2531, opened 2024-12-26
Fix this source: Royal Road watermark is back


Xampa5:
**App Version**:   3.7.4

## Describe this issue

Royal Road's 'invisible' text watermark is back. 

I download novels to read them offline on my oled smartphone screen and this permanent nagging in the middle of every chapter is infuriating and gets me out of the story.

I know it's supposed to be handled around line 80 of royalroad.py, but not how to fix it

`if style.endswith("{display:none;speak:never;}"):
                classnames.append(style[1:-27])`

EDIT: I've been able to fix it in post using Calibre editor. At least for the 8 novels I downloaded today, the only time there's a paragraph followed by a div, it's the nagging watermark. Which can be found with the regex
`<\/p><div>.*<\/div>`

Ctrl+F, Find: `<\/p><div>.*<\/div>`, Replace: `</p>`, Mode: Regex, All text files, click Replace all.

zGadli:
Close issue. The issue will be resolved in the next update.

PR #2532, opened 2024-12-29
Fixed royal road watermark issue. #2531

zGadli:


Xampa5:
I suspect deleting every <div> across Royal Road will lead to incomplete scrapping on some novels, as Royal Road authors are allowed to use HTML tags.
Example: https://www.royalroad.com/fiction/25225/delve/chapter/368068/004-statistics

zGadli:
@Xampa5 Didn't know that, I'll fix that in a second.

zGadli:
@Xampa5 I brute-forced it. I'll code a better approach in my free time.

PR #2540, opened 2025-01-05
Add Fenrir Translations source

kardigun:


dipu-bd:
The template looks very similar to `MadaraTemplate`. I have merged this since it is working, but can you check if using the template works.


```
from lncrawl.templates.madara import MadaraTemplate

class FenrirTranslationsCrawler(MadaraTemplate):
    base_url = ["https://fenrirtranslations.com/"]
```

kardigun:
You're right, this could be implemented based on `MadaraTemplate` with some adjustment for chapters selection to skip paid chapters (or actually chapters linking to the novel page itself).
Please tell me if using the template is preferred, I would then create another PR to fix.

dipu-bd:
yes, using a template is preferred. it is easier for maintenance

PR #2541, opened 2025-01-05
Add asianovel.net and wuxiasky.net sources

kardigun:


sources/en/w/wuxiasky.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from sources.en.a.asianovel_net import AsiaNovelNetCrawler
```
dipu-bd, line 4: it is better to not depend on a file inside the sources folder. for auto updates, it will create an issue. you can put this link inside the AsiaNovelNetCrawler itself. i.e.

```
class AsiaNovelNetCrawler(SearchableSoupTemplate, ChapterOnlySoupTemplate):
    base_url = [
        "https://www.asianovel.net/",
        "https://www.wuxiasky.net",
    ]
```

dipu-bd: or, if this template is a common one, we can put it inside `lncrawl/templates` as well

PR #2542, opened 2025-01-06
Add raeitranslations.com source

kardigun:


sources/en/r/raeitranslations.py
```diff
@@ -0,0 +1,54 @@
...
+import logging
+from urllib.parse import urlencode
+
+from lncrawl.core.crawler import Crawler
+from lncrawl.models import Chapter
+
+logger = logging.getLogger(__name__)
+
+
+class MyCrawlerName(Crawler):
```
dipu-bd, line 11: put a crawler name. i.e: `RaeiTranslationsCrawler`

sources/en/r/raeitranslations.py
```diff
@@ -0,0 +1,54 @@
...
+        "Accept": "*/*",
+        "Origin": "https://raeitranslations.com",
+        "Referer": "https://raeitranslations.com/"
+    }
+
+    def read_novel_info(self) -> None:
+        title = self.novel_url.rstrip("/").split("/")[-1]
+        novel_json = self.get_json(
+            f"{self.novel_json_url_prefix}{title}",
+            headers=self.json_request_headers
```
dipu-bd, line 27: headers are already added inside scraper. do you need to provide them separately?
check: https://github.com/dipu-bd/lightnovel-crawler/blob/6f1db35d8abb01538ac55e92272fa3a85a8e5910/lncrawl/core/scraper.py#L104-L105

Issue #2562, opened 2025-02-14
Fix this source webnovel.com


DomID00:
## Let us know

**Novel URL**:     https://www.webnovel.com/book/naruto-manipulation_31615358508137505
**App Location**:    EXE

**App Version**:   3.7.5
## Describe this issue

? Enter novel page url or query novel: https://www.webnovel.com/book/naruto-manipulation_31615358508137505
Retrieving novel info...
Exception in thread Thread-2 (read_novel_info):
Traceback (most recent call last):
  File "threading.py", line 1016, in _bootstrap_inner
  File "threading.py", line 953, in run
  File "lncrawl\templates\browser\basic.py", line 88, in read_novel_info
  File "C:\Users\MTURPI~1\AppData\Local\Temp\_MEI49362\sources\en\w\webnovel.py", line 69, in read_novel_info_in_soup
    self.get_csrf()
  File "C:\Users\MTURPI~1\AppData\Local\Temp\_MEI49362\sources\en\w\webnovel.py", line 36, in get_csrf
    self.csrf = self.cookies["_csrfToken"]
KeyError: '_csrfToken'

 ! Error: No chapters found
<class 'Exception'>
File "lncrawl\bots\console\integration.py", line 109, in start
    raise e
  File "lncrawl\bots\console\integration.py", line 103, in start
    _download_novel()
  File "lncrawl\bots\console\integration.py", line 87, in _download_novel
    self.app.get_novel_info()
  File "lncrawl\core\app.py", line 138, in get_novel_info
    raise Exception("No chapters found")

Issue #2514, opened 2024-12-04
Fix this source: Webnovel.com cant load chapter list


zillafox:
## Let us know

**Novel URL**:    [ <your novel url or query> ](https://www.webnovel.com/book/classroom-of-the-elite-year-2_24397792706417105)
**App Location**:  EXE
**App Version**:  NEWEST 3.72 

## Describe this issue

no chapter can be found

![Capture](https://github.com/user-attachments/assets/f65eda6b-9a11-4e42-9ac8-e345d22337b2)

Issue #2459, opened 2024-09-11
Error _csrfToken webnovel.com


YSaitOz:
-> Press  Ctrl + C  to exit

? Enter novel page url or query novel: https://www.webnovel.com/book/i'm-bombarding-the-supernatural-world-with-my-fire
arms_29917267208875605
Retrieving novel info...
Exception in thread Thread-1 (read_novel_info):
Traceback (most recent call last):
  File "threading.py", line 1016, in _bootstrap_inner
  File "threading.py", line 953, in run
  File "lncrawl\templates\browser\basic.py", line 88, in read_novel_info
  File "C:\Users\Allan\AppData\Local\Temp\_MEI128442\sources\en\w\webnovel.py", line 69, in read_novel_info_in_scraper
    self.get_csrf()
  File "C:\Users\Allan\AppData\Local\Temp\_MEI128442\sources\en\w\webnovel.py", line 36, in get_csrf
    self.csrf = self.cookies["_csrfToken"]
KeyError: '_csrfToken'

 ! Error: No chapters found
<class 'Exception'>
File "lncrawl\bots\console\integration.py", line 107, in start
    raise e
  File "lncrawl\bots\console\integration.py", line 101, in start
    _download_novel()
  File "lncrawl\bots\console\integration.py", line 85, in _download_novel
    self.app.get_novel_info()
  File "lncrawl\core\app.py", line 137, in get_novel_info
    raise Exception("No chapters found")

Press ENTER to exit...

YSaitOz:
Does anyone know how to resolve this error?

Issue #2402, opened 2024-06-19
www.webnovel.com _csrfToken


nonebutblank:
![image_2024-06-19_173109519](https://github.com/dipu-bd/lightnovel-crawler/assets/173296311/8c902697-e905-4a34-afa6-0bd719a0e7f0)

lesh558:
I'm having the same issue here, while accessing the website at the same time the exe program is running the website

PratikParmar810:
I think i have solved the issue of _crsfToken but it is not giving an output of epub file

YSaitOz:
> I think i have solved the issue of _crsfToken but it is not giving an output of epub file

How did you fix it?

PratikParmar810:
> > I think i have solved the issue of _crsfToken but it is not giving an output of the epub file
> 
> How did you fix it?

It has solved the "_crfToken" issue, but epub or any other files are not generated. It is another issue with Webnovel.com

PratikParmar810:
> > I think i have solved the issue of _crsfToken but it is not giving an output of epub file
> 
> How did you fix it?

https://github.com/dipu-bd/lightnovel-crawler/discussions/2445

PR #2572, opened 2025-02-23
Fix webnovel.com source

ihazpink:
The issue with chapter downloads was that webnovel stopped setting the `_csfrToken` cookie on requests to index page. My workaround is to request `www.webnovel.com/stories/novel` instead, where they still do set it. This could've been any page, I chose `stories/novel` at random.

image

Next, the issue with cover downloads was that they changed their URLs from `img.webnovel.com` to `book-pic.webnovel.com`. They also changed the URL parameters and format slightly. My (hacky) fix works - the cover downloaded for me - but I'm not sure whether it will work for all cases. In particular I didn't bother to find the `imageId` parameter after I saw that it works without it, and I kept the `coverUpdateTime` parameter even though it's probably not used anymore.

image

This PR should address:
#2562 
#2514
#2459
#2402

### DIFF
--- a/sources/en/w/webnovel.py
+++ b/sources/en/w/webnovel.py
@@ -32,7 +32,7 @@ class WebnovelCrawler(BasicBrowserTemplate):
 
     def get_csrf(self):
         logger.info("Getting CSRF Token")
-        self.get_response(self.home_url)
+        self.get_response(f"{self.home_url}stories/novel")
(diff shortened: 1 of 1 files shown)


dipu-bd (approved):

PratikParmar810:
how do we add this in lncrawl?? So we can add our own _crftoken and webnovel_uuid

PR #2576, opened 2025-02-25
Site was shutdown OR name was changed

Magos-Technicus:
I put the entire thing as a comment because of it
hope the site is still running in a internet corner somewhere with a different name


dipu-bd:
you can add this site to rejected.json to disable it

Magos-Technicus:
i'll do it then, thanks

PR #2595, opened 2025-03-18
Fix NovelNext

redowan99:


sources/en/n/novelnext.py
```diff
@@ -24,3 +27,7 @@ def initialize(self) -> None:
                 "strong": r"NovelNext\.com",
             }
         )
+
+    def select_chapter_body(self, soup: BeautifulSoup) -> Tag:
+        sleep(5)
```
dipu-bd, line 32: instead of doing sleep here, configure rate limiting in initialize function:

```
    def initialize(self) -> None:
        self.init_executor(ratelimit=0.2)
```

for docs check: `lncrawl/core/taskman.py`

dipu-bd (changes requested):

Issue #2598, opened 2025-03-23
https://arcanetranslations.com


AmazingJayy:
- Language: English
- Supports Searching: Yes
- Contains Machine Translations: No
- Contains Manga/Manhua/Manhwa: No
- Has CloudFlare Protection: Yes

PR #2636, opened 2025-05-21
added source arcane translation

jere344:
#2598
Not sure how to deal with locked chapter, maybe I should have ignored it, then we can just delete the select_chapter_body override I added.
I raised an error so that it doesn't create the chapter file and set it as success:true in meta.json. So if we run the app again, it will fill in the missing chapters instead of leaving it as "Login to buy access to this content" despite the chapter being available since.


dipu-bd:
looks good

PR #2638, opened 2025-05-22
added tags and summary to templates + new source novlov

jere344:


dipu-bd (changes requested):
Thanks! This was needed to be done.

PR #2646, opened 2025-05-26
Automatic audiobook (.mp3) generation with edge-tts

IanRockwell:
This pull request adds support for automatically generating audiobooks using [edge-tts](https://github.com/rany2/edge-tts) — a Python module that enables access to Microsoft Edge's online text-to-speech service without any authentication or needing Edge and/or Windows installed

I thought this would be a great addition to the project, especially for situations where I can't always be looking at the screen but still want to enjoy a light novel hands-free.

This is my first ever pull request, so if I’ve made any mistakes or overlooked something, please don’t hesitate to point it out. And if you feel this feature is outside the scope of the project and prefer not to include it, that’s completely understandable as well.


dipu-bd:
Thank you for the pull request. 

Since all good epub readers provides a TTS functionality, I don't see how this will bring much value. I usually use Moon Reader+ for TTS. It also has a lot of voice types and speed controls to choose from as well. 

Generating mp3 files and downloading them from an online tool can be often unreliable and quite slow. Also, there is a risk of getting your IP blocked when transforming a novel with 3000+ chapters.

However, we can still have mp3 output as an optional choice. I like that the `edge-tts` package is quite small and does not have a lot of dependencies.

Issue #2483, opened 2024-10-12
App unable to handle multiple user while crawling novel


ruchitaza7:
While running the telegram bot, it is unable to handle multiple users at the same time. When one session is active, it doesn't allow other users to start a session until the first session is fully completed and closed.

PR #2706, opened 2025-07-07
rework of telegram bot

CryZFix:
1. replaced deprecated method compress_books
2. reworked logic for parallel work #2483


dipu-bd:
can you fix the linting issues?

Issue #2727, opened 2025-08-01
[Fix novel543] Second Page of Each Chapter Is Not Scraped


Mayonnaisu:
## Let us know
Yeah, I have searched the existing issues, and found nothing related.

**Novel URL**: [https://www.novel543.com/0517254464/](https://www.novel543.com/0517254464/)
**App Location**: EXE

**App Version**: 3.10.1

## Describe this issue

There is no error. But when I read the downloaded novel, I felt something is missing. When I checked the novel directly on the website, it turns out that each chapter is split into 2 parts on different pages. And the program only scrapes the first page. Honestly, I tried to modify the source myself, but I failed to do so after I can't figure out where should I concatenate "_2" to each chapter url after duplicating each of them. My programming level is just way too low. So I have to trouble someone more capable with this. Thanks in advance!

PR #2730, opened 2025-08-02
fix novel543.py multi page fetch #2727

jere344:
#2727

### DIFF
--- a/sources/zh/novel543.py
+++ b/sources/zh/novel543.py
@@ -146,4 +146,20 @@ class Novel543(Crawler):
         if tip_div and "溫馨提示" in tip_div.text:
             tip_div.decompose()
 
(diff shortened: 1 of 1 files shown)


jere344:
Can you give me the novel that failed and a chapter with missing content ?
It works for thoses I tried but there may be other cases

jere344:
Hmm if you're running it through pip I have no idea, maybe it's not really hitting the file and there's some cache-

You can try putting a simple ` raise Exception("text exception")` in the read_novel_info and run it again, just to check it's actually running the right file. 
image

jere344:
If it is not running this file, then you should maybe run the app from source instead
If it is running this file, then it's weird- make sure you deleted the novel folder so that it properly re download it

PR #2745, opened 2025-08-14
Added comments to multiple files inside \lncrawl\.

zGadli:
Used cursor with the GPT-5 model to generate helpful comments. I feel like this project is bloated, and It could probably be broken down into multiple repos.


dipu-bd:
Breaking it down will take some extra maintenance time.

lncrawl/core/crawler.py
```diff
@@ -33,13 +39,10 @@ def __init__(
...
-        """
-        Creates a standalone Crawler instance.
+        """Create a standalone crawler instance.
 
-        Args:
-        - workers (int, optional): Number of concurrent workers to expect. Default: 10.
-        - parser (Optional[str], optional): Desirable features of the parser. This can be the name of a specific parser
-            ("lxml", "lxml-xml", "html.parser", or "html5lib") or it may be the type of markup to be used ("html", "html5", "xml").
+        - workers: Expected concurrency for downloads
+        - parser: Desired BS4 parser or markup type
```
dipu-bd, line 45: it is removing some useful docs

dipu-bd:
This change removes all the manually created docs. Some of those docs contained useful information. I would prefer manual docs to remain unchanged, or rewritten without removing any information.

PR #2777, opened 2025-09-13
Add rate limit for downloads

gumulka:
Some websites limit the number of accesses to it and it should also be in the interesst of all to be good netizens and not overly stress pages.

Allow the user to restict the access to pages to x per minute, whereas values smaller then 1 result in less then one access per minute. 0.5 will result in one access every two minutes, reducing strain on the webservers we get our novels from.

Default has been set to 1 per second, which is quite high, but in the spirit of the previous implementation which did not throttle at all.

### DIFF
--- a/lncrawl/core/arguments.py
+++ b/lncrawl/core/arguments.py
@@ -303,6 +303,14 @@ _builder = Args(
             help="Do not prompt to close at the end for windows platforms.",
         ),
         Args(
+            "--ratelimit",
+            dest="ratelimit",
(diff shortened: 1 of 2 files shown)


dipu-bd:
we already have a rate limiting

augustanational:
> we already have a rate limiting feature: https://github.com/dipu-bd/lightnovel-crawler/blob/dev/lncrawl/core/taskman.py

@dipu-bd Is there any easy way to use this feature?

PR #2858, opened 2026-01-27
Contribution Improvements + Build System Upgrades + Docker Arm Support

VibrantClouds:
This one was a little bit of a doozy, this pull request deprecates #2840 in favor of a more comprehensive set of adjustments.

Given that there are a chunk of CICD changes to this pull request, please don't hesitate to reach out with questions in regards to the 'why'. Quite frankly, there was a lot of problems getting this thing to work right, it wasn't very friendly to forks which I presume isn't much of an issue when most of your pull requests are simply users adjusting source scrape rules.

For full transparency, Claude Code was utilized to a reasonable degree helping with some of the legwork with this project. I **do** work in the industry and am very familiar with all of the technologies used in this project, I've vetted everything thoroughly. (Un)fortunately CICD is one of the things I've put a lot of time into over the years. (Granted all of that work is in Azure Pipelines, though Github Actions are similar-ish enough!)

### Architecture Improvements
- Docker builds now additionally target ARM systems
### Document and Onboarding
- Added a contribution guide, architecture overview, crawler creation guide and docker documentation
- Added a CLAUDE.md file for AI-assisted development
  - I figure instead of getting swarmed with issues requesting features, with the proper upfront context, users not very used to Python or scraping may be able to accomplish a chunk of legwork themselves.
  - Especially for a project as simple as this, WebFetch tools and a LLM which can read JSON can absolutely determine parsing rules without much issue
- Added a forking guide to help downstream maintainers understand how the Github actions work, and how to manage their own environments
### CI/CD Updates
- Custom workflow run names for easier identification
- Relaxed version tag enforcement on forks (allows forks to run release workflows, the enforcement is retained for the upstream lncrawl repository as these publishes are what the public will be using)
- Updated the base image reference to support forked repositories
- Expanded the build matrix to include Python 3.14 and 3.15 during validation

### Dependencies and Build Fixes
- Switched to UV where possible for faster dependency resolution
- Fixed the wheel installation test command (Previous version was referencing the old CLI syntax, this properly uses the new one)
- Fixed Pydantic v2 compatibility (`encoded_string()` is not supported on really old python versions, `str(...)` wrapping is a standard method across language versions
- Fixed rich text escaping for file paths containing brackets 

## Validation
Reference the following Github builds for proof of completeness and passing all stages:
- Build Base Image: https://github.com/VibrantClouds/lightnovel-crawler/actions/runs/21378115381
- Build and Publish Release: https://github.com/VibrantClouds/lightnovel-crawler/actions/runs/21378115411

PR #2860, opened 2026-01-28
Python Compatibility Tweaks

VibrantClouds:
Partial update of #2858 

Bracket string escapes added to `utils.py` solves `MarkupError` during pipeline builds, see: https://github.com/VibrantClouds/lightnovel-crawler/actions/runs/21377160806

Switches `encoded_string()` usage to more standardized `str(x)` syntax for backwards compatibility in older python versions. This works on recent versions, but as you're including some older versions in the version matrix, this consistently works.  see: https://github.com/VibrantClouds/lightnovel-crawler/actions/runs/21377572883/job/61537208423

`encoded_string()` may be valid if you are willing to drop certain versions of Python's support. (Version pinning to Pydantic 1.x may be viable?)

Through testing, `str()` works just fine here though!

### DIFF
--- a/lncrawl/services/sources/utils.py
+++ b/lncrawl/services/sources/utils.py
@@ -100,12 +100,12 @@ def import_crawlers(file: Path) -> Generator[Type[Crawler], None, None]:
         mod_name = hashlib.md5(file.name.encode()).hexdigest()
         spec = importlib.util.spec_from_file_location(mod_name, file)
         if not (spec and spec.loader):
-            logger.info(f"[{file}] Unexpected spec")
+            logger.info(f"\\[{file}] Unexpected spec")
             return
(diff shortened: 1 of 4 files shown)


lncrawl/services/sources/utils.py
```diff
@@ -136,7 +136,7 @@ def import_crawlers(file: Path) -> Generator[Type[Crawler], None, None]:
         urls = [str(url).lower().strip("/") + "/" for url in urls]
         urls = [url for url in set(urls) if validate_url(url)]
         if not urls:
-            logger.info(f"[{file}] No base url: {crawler}")
+            logger.info(f"\\[{file}] No base url: {crawler}")
```
dipu-bd: why backslash was added here?

PR #2876, opened 2026-02-26
fix: localhost for dynamic local IP on web UI for backend API requests

dr-dao:


dipu-bd:
ops, commited my local ip

PR #2924, opened 2026-04-14
Update 2026_03_19_21_33_54_artifact_file_size_column.py

alzamer2:
this fix when code try to create Column file_size, which is already exsit when it try to migrations 2026_03_19_21_33_54_artifact_file_size_column.py


dipu-bd:
the column should not exists if all previous migrations were done correctly. i suggest dropping the migration and run it again.

```
# downgrades last migration
> lncrawl dev migrate down

# applies migrations
> lncrawl dev migrate up
```

alzamer2:
hmm i will try it out
but i think the migrations  logic is faulty, as it should do it with order to newer

dipu-bd:
Should be working in the new version

PR #2940, opened 2026-04-27
update etc\wuxiaworld.com

alzamer2:
with this update, you can used dumped typescript files (*.ts) to automatically generate proto files then convert it to JSON descriptor for wuxiaworldcom.py

it include all  wuxiaworld API requests

ps: you can remove any unnecessary from wuxiaworld_typescript before you run the scripts, but make sure what you removed is not dependency for what you need


alzamer2:
ps2: you need to https://github.com/dipu-bd/pyease-grpc/pull/9 so you can support custom fieldoption

dipu-bd:
you don't need to commit 87 files to generate a proto! i would rather keep them locally or create a script to fetch them for proto file generation

alzamer2:
> you don't need to commit 87 files to generate a proto! i would rather keep them locally or create a script to fetch them for proto file generation

if you can create script to scrap them, that will be welcomed, i tried and i failed, that why i scraped them manually

dipu-bd:
I am going to close this PR since there is no plan to merge it in current state

PR #2942, opened 2026-05-01
Fixed Webview Not Starting on Windows + Makefile's Windows UV Path

FiredAnt8442448:
Added ctx.setup() before ctx.users.get_admin() so that the database is finished bootstrapping before the users table gets queried. Also fixed the makefile's windows UV path so that it actually works when building on windows.


lncrawl/server/webview.py
```diff
@@ -13,6 +13,8 @@
 
 def start() -> None:
     host = "127.0.0.1"
+    
+    ctx.setup()
```
dipu-bd: this is already fixed

PR #2946, opened 2026-05-06
Added source novelfrance.fr

biscotteman:
Added source for https://novelfrance.fr/, previously known as massnovel.fr

I'm not sure if I need to modify the README.md manually to add the source?

### DIFF
--- a/sources/fr/novelfrance.py
+++ b/sources/fr/novelfrance.py
@@ -9,52 +9,51 @@ from lncrawl.core import Chapter, LegacyCrawler
 logger = logging.getLogger(__name__)
 
 
-# We have a lovely API for this one, so let's use it! To the devs at NovelFrance, thank you! <3
 class NovelFrance(LegacyCrawler):
     base_url = ["https://novelfrance.fr"]
     has_manga = False
     has_mtl = False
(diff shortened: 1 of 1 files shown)


dipu-bd:
Please remove unnecessary comments. Snake case is preferred over camel case. i.e. use `has_more` instead of `hasMore`

biscotteman:
Hi :)
As requested, I've changed the variables from camel case to snake case and removed some unnecessary comments
I'm not sure about *how many* comments I should remove (I usually like my code to be as understandable as possible), so if you feel like this is still too much, I'll make another pass :)

dipu-bd:
Having comments is certainly good practice. But the problem is that nobody maintains the crawlers they have committed. I have no bandwidth to fix comments later on. As crawlers are supposed to be straightforward and simple, the code itself should do the talking rather than comments.

dipu-bd:
Having comments is certainly good practice. But the problem is that nobody maintains the crawlers they have committed. I have no bandwidth to fix comments later on. As crawlers are supposed to be straightforward and simple, the code itself should do the talking rather than comments.

biscotteman:
Alright, comments removed then :)

PR #2988, opened 2026-05-21
Fix for NovelFire.net

TheMr-Fool:
Added logic duplicate title from chapter content.


sources/en/n/novelfire.py
```diff
@@ -55,4 +55,10 @@ def read_novel_info(self) -> None:
     def download_chapter_body(self, chapter) -> str:
         soup = self.get_soup(chapter["url"])
         contents = soup.select_one("div#content")
+
+        # Remove the duplicate chapter title at the top (a <p> containing a <strong> tag)
+        first_p = contents.find("p")
+        if first_p and first_p.find("strong"):
+            first_p.decompose()
```
dipu-bd: if some novel doesn't have a title it'll remove first paragraph.

dipu-bd (changes requested):

PR #2993, opened 2026-05-22
Fixes Epub.py and NovelFire.py

TheMr-Fool:
Removed the # serial number heading from chapters that already have a number in their title. Specifically:
The original code always added <h4 style="opacity: 0.8">#{chapter.serial}</h4> above every chapter title. Changeged it so it only adds that line if the chapter title contains no numbers. So:

"Chapter 1 Sunny" → has a number → no #1 added
"Sunny" → no number → #1 gets added

That way sites where the title already includes the chapter number won't get the duplicate #1, but sites where the title is just a plain name still get the serial number shown.

Actual fix to NovelFire now removes duplicate titles (won't work if the title repeats 3 times(tested); the last code only worked for some


lncrawl/services/binder/epub.py
```diff
@@ -100,11 +100,19 @@ def build_chapter(chapter: Chapter) -> epub.EpubHtml:
         text = ctx.files.load_text(chapter.content_file)
     else:
         text = "<p><em>No content available</em></p>"
+
+    # Only show #serial if the title doesn't already contain a number
+    serial_heading = (
+        ""
+        if re.search(r"\d", chapter.title)
+        else f'<h4 style="opacity: 0.8">#{chapter.serial}</h4>'
```
dipu-bd: you can set `aria-hidden="true"` attribute in h4 intead of hiding it. since your case is just to not let tts speak it out loud.

dipu-bd (changes requested):

lncrawl/services/binder/epub.py
TheMr-Fool: serial_heading = (
    ""
    if re.search(r"\d", chapter.title)
    else f'<h4 aria-hidden="true">#{chapter.serial}</h4>'
)

dipu-bd: i want to show the serial even if chapter title contains it. otherwise this breaks consistency. some chapter will have the serial, some won't.

dipu-bd (changes requested):

lncrawl/services/binder/epub.py
TheMr-Fool: Got ya that's fine. I can always just edit them out on my on

dipu-bd: This has been stale for a while. As I have requested, keep the serial but put `aria-hidden="true"` attribute in the h4 tag.

Or, you can remove changes to the file, and we can merge the novelfire fix

TheMr-Fool: done

TheMr-Fool (commented):
done

PR #3009, opened 2026-05-31
Add search for novelfire

GabrielCWT:


GabrielCWT:
Unfortunately I am unsure how to go about testing this feature. 

I have ran `make index-gen` to update the index and then ran `uv run python3 -m lncrawl search "Barbarian" -s "novelfire"`

However, it still errors with `NotImplementedError()`. Would love to get some help on this!

dipu-bd:
It needs index generation. If I merge this, it should start working.

`search` feature is not yet supported with the new architecture properly. once other priority features are done, it will be picked up.

PR #3070, opened 2026-07-01
handle the throttle and text obfuscation for novelhi.com

zimorok:
## What does this PR do?
using the site API token endpoint but will cost more time to download and is not suitable for multi user download unless you activate the proxy. multiple concurrent download is limited and the site will throttle you from 15s to 180s.   this update will try to handle the throttle by waiting within the throttle delay and keep retrying

## Type of change

- [x] Source Crawler
- [ ] New source
- [ ] Feature / enhancement
- [ ] Docs / chore

## Details
https://novelhi.com/novel/comedy/scholars-advanced-technological-system

## known error
the site will throttle your download. lncrawl may fail sometime to download the novel, so far, there is no way to 'repair' the failed downloaded content

### DIFF
--- a/sources/en/n/novelhi.py
+++ b/sources/en/n/novelhi.py
@@ -1,19 +1,142 @@
 # -*- coding: utf-8 -*-
+import codecs
 import logging
+import re
+import time
 from urllib.parse import quote
 
-from lncrawl.core import Chapter, LegacyCrawler, Volume
+from bs4.element import NavigableString
+
+from lncrawl.core import Chapter, LegacyCrawler, PageSoup, Volume
+from lncrawl.exceptions import LNException
 
 logger = logging.getLogger(__name__)
 search_url = "%s/book/searchByPageInShelf?curr=1&limit=10&keyword=%s"
 fetch_chapter_list_url = "%s/book/queryIndexList?bookId=%s&curr=1&limit=50000"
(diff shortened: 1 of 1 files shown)


dipu-bd (changes requested):
Thanks for tackling novelhi.com's throttling and font obfuscation — this is clearly a hard site to scrape reliably.

A few things I'd like addressed before merging:

1. **The retry logic doesn't cover the requests most likely to need it.** `read_novel_info` and `search_novel` still use the plain `get_soup`/`get_json` calls, so the very first request for a novel (where a 429 is most likely) fails immediately with no retry — the opposite of what this PR is meant to fix.
2. **`_get_response_with_retry` has mismatched retry constants** (`range(browser_max_retries)` paired with a break check against `max_retries - 1`), so it always sleeps a wasted ~120s on the final attempt before giving up, and effectively retries one fewer time than intended.
3. **The new search URL builder is fragile.** It replaces the previously-reliable `item["simpleName"]` link with a guessed `/novel/<genre>/<slug>` built from a hardcoded genre map and re-slugified title — any unmapped genre or slug mismatch silently breaks search.
4. **The legacy fallback chapter-extraction path never de-obfuscates text**, so a chapter that lands in that branch while obfuscated will produce rot13 garbage in the output.

Given the size of the change (~350 new lines in one file) and the above, I think this needs another pass — particularly consolidating the two parallel retry/backoff implementations and making sure the throttle handling actually covers the novel-info/search requests. Happy to take another look once those are addressed.

PR #3105, opened 2026-07-13
Add CLI resume and crawl rate limit options

HaProxyMage:
## What does this PR do?

Add CLI resume and crawl rate limit options

## Type of change

- [ ] Bug fix
- [ ] New source
- [x] Feature / enhancement
- [ ] Docs / chore

## Details

Add --resume / --missing to the CLI crawl command to download only chapters that are not marked as downloaded.
Add --refresh to force re-downloading selected chapters.
Add --rate-limit to override the crawler task rate for a single CLI crawl run.

The crawler services already skip available chapters and the scheduler already supports fetching missing chapters, but the non-interactive CLI did not expose that behavior clearly. This makes interrupted downloads easier to continue and gives users a per-run throttle without editing crawler source code.

## Checklist

- [x] `make lint` passes with no errors
- [x] Tested manually (crawl command or server)


lncrawl/commands/crawl/crawl.py
```diff
@@ -33,6 +33,24 @@ def crawl(
         metavar="N",
         help="Download latest few chapters",
     ),
+    resume: bool = typer.Option(
```
dipu-bd, line 36: Why add an explicit `resume` since the default behavior of the crawl command is to resume the download?

dipu-bd (approved):
Overall looks good. Thanks for taking up this task. I will wait for further clarification on the `resume` option before merging it

lncrawl/commands/crawl/crawl.py
dipu-bd: ah I understand, you are not even fetching the list of chapters that is marked with `is_crawled` True

HaProxyMage: Exactly. That keeps the progress bar and queued task count focused on only missing chapters, which is especially useful for large novels or interrupted downloads. The default behavior remains unchanged.

PR #3140, opened 2026-08-03
Fix bug on scrap NovelPhoenix, now can detect all existing pagination.

BaruchManiac:
…from the NovelPhoenix site.

It can now detect all existing pagination and navigate through the pages to retrieve every chapter.

## What does this PR do?

Fixed the crawler, which previously only read the first 100 chapters from the NovelPhoenix site.

It can now detect all existing pagination and navigate through the pages to retrieve every chapter.

## Type of change

- [X] Bug fix
- [ ] New source
- [ ] Feature / enhancement
- [ ] Docs / chore

## Checklist

- [X] `make lint` passes with no errors
- [X] Tested manually (crawl command or server)


.python-version
```diff
@@ -1 +1 @@
-3.9
+3.12
```
dipu-bd: please remove this file from the PR

dipu-bd (approved):

Issue #3148, opened 2026-08-06
Fenrir hidden text


dipu-bd:
There's a long gibberish text hidden by the web between sentences, when extracted it make the epub impossible to read.

watzeedzad:
[https://github.com/lncrawl/lightnovel-crawler/pull/3180](url)

PR #3180, opened 2026-09-13
Fix download body and filter only free chapter for Ferrireaml

watzeedzad:
## What does this PR do?

Filter only free chapter in ferireaml source and correct its download method to filter out unwanted tag.

## Type of change

- [x] Bug fix
- [ ] New source
- [ ] Feature / enhancement
- [ ] Docs / chore

## Details

* The old chapter fetching logic does not filter paid chapter and cause crawler to crawl paid chapter that always failed.
* Update the download chapter method to remove invisible tag that contains text seem to be identifier for the site.

## Checklist

- [x] `make lint` passes with no errors
- [x] Tested manually (crawl command or server)


watzeedzad:
Related to #3148

